PR #48, opened 2023-12-07
audio player first pass

webofpies:
First pass for audio player

There's a temporary "browse" button for adding an audio file to the player. This button won't be there for the final version probably.

What needs to be implemented on back end:

1) storing bookmark positions (from 1 to 99 probably) for later retrieving
2) storing last play position (from 1 to 99 probably) so user can resume immediately when reopening a book. Variable "lastPlayTime" is already declared and used in the script.

image


jzohrab:
Super, thanks very much!

The PR touches a few lines of code it shouldn't, looks like your editor was auto-reformatting some files.  In some cases it's good.  I'll try working with the commit in a feature branch.

PR #51, opened 2023-12-08
 Add Sudachi Parser for parsing japanese without MeCab

fanyingfx:


fanyingfx:
Add the `SudachiParser` class in the` sudachi_parser.py` , as a replacement of `JapaneseParser`which requires MeCab, has the same methods with `JapaneseParser`, and change the parser in the` registry.py`.
For the testing issues:
  one test failed in the unit test:

image

This is because the Sudachi have the parsed result with MeCab, and the dictionary is different.

one test failed in the acceptance test:
image

This possibly because  the default `is_supported()` `return True` which is inherented from `AbstractParser`

And most of the `get_reading`() code are just copy from ` mecab_parser.py`, these code are duplicated.

fanyingfx:
Find a solution to use MeCab on Windows 64bit, seems not need to add another Japanese parser.

PR #52, opened 2023-12-08
fixes #33 - set a term(s) status via the keyboard

### LINKED ISSUE #33: Hotkey arrow up and down to increase/decrease status number
Hotkey left-right moves to terms, so up/down could change status.

Would need to go up related to the current status.  If multiple terms chosen, could just start with the lowest status.  Go from 1-2-3-4-5-WellKnown, skip "ignored".

The function to update is in lute.js, `handle_keydown` -- at least, that is how I was intending to do it.  If there is a better option, LMK.

robby1066:
This PR makes the following changes to /static/js/lute.js:

### Adds a new function called `increment_status_for_marked_elements`
This function collects selected term elements, groups them by their `data_status_class` level, then submits them in batches to be updated with `update_status_for_elements`.

This function accepts a single argument `shiftBy`, an integer indicating the amount and direction to increment the status class as defined in the constant `validStatuses`.

### Abstracts the code in `update_status_for_marked_elements` into a new generic function `update_status_for_elements`
Separating the selection of elements from the code that submits them to the server allows other functions to update statuses for different use cases. 

`update_status_for_elements` takes two arguments
- `new_status`: the status that will be set on the elements
- `elements`: a nodeList, array, or jQuery selection of elements to update

### Passes a new argument to `update_selected_statuses`
`update_selected_statuses` was doing a redundant selection of elements to update what was sent the server by `update_status_for_marked_elements`. Passing an `elements` argument instead allows more flexibility upstream and removes duplicated element selection code.

### Binds arrow up and arrow down keys to `increment_status_for_marked_elements` in `handle_keydown`

### DIFF
--- a/lute/static/js/lute.js
+++ b/lute/static/js/lute.js
@@ -12,7 +12,6 @@ let LUTE_CURR_TERM_DATA_ORDER = -1;  // initially not set.
  */ 
 let LUTE_HOVERING = true;
 
-
 /**
  * When the reading pane is first loaded, it is set in "hover mode",
  * meaning that when the user hovers over a word, that word becomes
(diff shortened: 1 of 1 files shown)


jzohrab:
Had to make one small change: setting the status to 0 isn't allowed, as that implies a missing term.  The model isn't clear on this point.

PR #63, opened 2023-12-20
Add Modern Chinese

fanyingfx:
Adding the Modern Chinese Parser using [hanlp](https://github.com/hankcs/HanLP/tree/master) (The English version document seems not good. 😞) which I think the parsed result is very nice.
It requires Pytorch, Tensorflow, so by default the is_supported() return False, until user installed the `hanlp` by pip.
It needs to download model, and the path is configurable . About model path configuration: https://hanlp.hankcs.com/docs/configure.html


fanyingfx:
Local branch rebased, this pr can not sync the new changes, will create a new PR

PR #64, opened 2023-12-20
Feature:  Text Options

webofpies:
- added text control options (for book text):
	- font size
	- line height
	- text container width
	- column count

misc:
- added title attributes to the player controls
- player gets to be sticky automatically if the section width cannot accommodate player width
- removed unused stuff from player.js and styles.css

Note: to open the options, a temporary button has been added to the header section

image


jzohrab:
Closing in favor of #68 .

PR #69, opened 2023-12-22
Add Modern Chinese Parser

fanyingfx:
Adding the Modern Chinese Parser using [hanlp](https://github.com/hankcs/HanLP/tree/master) (The English version document seems not good. 😞) which I think the parsed result is very nice.
It requires Pytorch, Tensorflow, so by default the is_supported() return False, until user installed the hanlp by pip.
It needs to download model, and the path is configurable . About model path configuration: https://hanlp.hankcs.com/docs/configure.html
If hanlp is not installed, using [pkuseg ](https://github.com/lancopku/pkuseg-python/blob/master/readme/readme_english.md) as default Chinese Parser instead.


jzohrab:
Hey @fanyingfx - this looks nice, and it looks like everything is there with tests, the check for support, etc.  Great work, thank you!

I only hesitate to merge this because of the extra requirements and the need for some kind of imported model, which requires documentation and a way to test this with and without the packages.

This is a different way of augmenting Lute than what I had originally had in mind architecturally -- really, I haven't come up with a good way to manage this kind of thing -- but perhaps your idea here (of pip installing extra packages) is in fact the right way to move forward.

Are you able to do record a video showing this working on your system, including any supporting files you might have installed?

Thanks again for this PR, it looks interesting!

jonathonmh:
I'm keen to know if this generalises for Cantonese as well. 
I believe I've seen a python library, for Cantonese, that didn't rely on an AIML solution, i.e. pytorch, tensorflow - i.e. used traditional techniques.

Notwithstanding that this PR might be a great solution, I intend to investigate separately, perhaps there's another solution.

fanyingfx:
> I believe I've seen a python library, for Cantonese, that didn't rely on an AIML solution, i.e. pytorch, tensorflow - i.e. used traditional techniques.
Hi, @jonathonmh, thanks for you advice, I find the [jieba ](https://github.com/fxsjy/jieba) is eaysier to install (it has no other dependencies).So I replace the default  text segmentation from pkuseg to the jieba. If user wants a up-to-date parser(also with up-to-date dictionary) can install the hanlp by pip, then lute will use hanlp to parse Cantonese.

fanyingfx:
Hi @jzohrab , I know your hesitation. The extra requirements like the model is a  annoying thing. I just change a bit of  my code, that using jieba as default parser(which hasno other dependencies). 
Yeah, I can record a video. It is a good idea that the results or issues? will be more clearly demonstrated.

fanyingfx:
When I recording a video about Chinese parser, I came across a problem. I found when set term text, the text will be parsed again.
```python
  @text.setter
    def text(self, textstring):
        ...
        tokens = lang.get_parsed_tokens(t)
        ...

```
And when set multiple term using hanlp as parser(Jieba encounters this kind of issue infrequently, but it frequently parsed multiple words term as single word term), the parsed_tokens() result may differ from  it in sentence. It will lead to some multiple words terms can not matched. 
For example:
image
in the sentence "竹条编" can be parsed to "竹条" and "编", when only parse "竹条编" , the result is "竹条编".

fanyingfx:
When try to solve the multiple terms words for Chinese parser, I have an idea: maybe the zero-width-space for multiple words term is not very suitable for Chinnese ?

jonathonmh:
Hi @fanyingfx

Looking at the jieba github is a little intimidating! Does this work for only Mandarin?

The Cantonese library I had seen is pycantonese, which also performs segmentation, and conversion to jyupting (i.e. pinyin).
(the conversion to jyutping could be used to prepopulate the ruby field as discussed https://github.com/jzohrab/lute-v3/issues/71)

There may need to be separate solutions for Manadarin vs Cantonese here.

fanyingfx:
Hi @jonathonmh , after some searching, it seems jieba does not support  Cantonese well  neither support the jyupting.
I think separate soulutions for Manadarin and Cantonese is needed.

fanyingfx:
Hi @jzohrab , I recorded a lute video about this PR. It only using jieba as parser ,for the hanlp , I find it's hard to make the multi-words term works, so it was removed.
https://github.com/jzohrab/lute-v3/assets/57335844/379e7cb8-f1ac-418e-bce1-35173db5d9b9

jzohrab:
Hey sorry haven't answered the various questions :-) -- it's getting busy!

> I found when set term text, the text will be parsed again.

Yes, that ensures that the term text has the zero-width token separators in the right locations.

> maybe the zero-width-space for multiple words term is not very suitable for Chinese ?

Maybe, but that's a big change to the parsing and text matching model.  The zero-width-space token separator _vastly_ simplifies all pattern matching, which is important.

> Yeah, I can record a video. It is a good idea that the results or issues? will be more clearly demonstrated.

I don't understand what you're trying to say :-)

> in the sentence "竹条编" can be parsed to "竹条" and "编", when only parse "竹条编" , the result is "竹条编".

Right, this can happen in any language, where a multi-word term covers other terms.  :-P  I think the right solution here (well, as right as I can come up with) is to show the "word components" as well for any multi-word terms in the popup, because sometimes that's useful or important.

> The extra requirements like the model is a annoying thing.

Maybe, but it could also be an elegant solution to a difficult problem!  Having a "plug-in" architecture that's extensible for certain needs is a nice way to solve things, and it avoids the mess of a single massive thing trying to handle every possible case.  As long as the management is clear, and the code can handle the component getting added and removed simply and cleanly, it could work great.  But honestly I'm not sure _how_ to keep the management clear.  Using pip means that the environment gets "polluted" with extra things -- hopefully that's not a problem, but I just don't know. :-)

Nice video, thank you!  It certainly looks like the right way to go.

fanyingfx:
> Hey sorry haven't answered the various questions :-) -- it's getting busy!

No problem, I know you are so busy,  just in the Discord I can see you in the everythere XD.

> Maybe, but that's a big change to the parsing and text matching model. The zero-width-space token separator vastly simplifies all pattern matching, which is important.

When I try to replace the zero-width-space method, things became more complicate :(.

> I don't understand what you're trying to say :-)

Don't mind. It's just  just talking to myself XD.

> Right, this can happen in any language, where a multi-word term covers other terms. :-P I think the right solution here (well, as right as I can come up with) is to show the "word components" as well for any multi-word terms in the popup, because sometimes that's useful or important.

If my undestanding is correct, about solution: the multi-word terms can also match the seperate "words componets" as a whole term? 

Another thing, just for the Chinese parser what I'm using, it support the user defined dictionary,  can make user defined terms has high priority for the parser, and easy to load. If add all multi-word terms to the user dicitionary, then parsed result can be deterministic. As far as I know, the Mecab support similar thing. Then there is another problem about the dictionary load/update/delete.

> Having a "plug-in" architecture that's extensible for certain needs is a nice way to solve things, and it avoids the mess of a single massive thing trying to handle every possible case. 

Yeah, a "plug-in" architechture is good, if we can define a clear and general process about how extra things download/load/delete ...etc, things will be easy and clean. I agree there is a soultion for this thing, but I don't know neither XD. Need time to think and try.

krzysztof-wilczynski:
Hey there, any update on this?

fanyingfx:
Hi @krzysztof-wilczynski ,  
After some tests, I have found that some additional code modifications are needed for  Chinese when using Lute. If you are interested, you can try my Mandarin fork https://github.com/fanyingfx/lute-v3

krzysztof-wilczynski:
> Hi @krzysztof-wilczynski , After some tests, I have found that some additional code modifications are needed for Chinese when using Lute. If you are interested, you can try my Mandarin fork https://github.com/fanyingfx/lute-v3

Your fork works perfectly, thanks.

jzohrab:
Closing in favor of #408.  The source branch has also changed to chinese_new in the Mandarin fork.  I've chatted briefly about the chinese_new branch with @fanyingfx , and hope to pull some of that work into Lute, so that Mandarin is part of this main release.  Thanks!

PR #70, opened 2023-12-22
Move completed books checkmark on left, line up titles

### LINKED ISSUE #57: Completed books checkmark on left, line up titles
Per https://discord.com/channels/1074759089051160647/1181103335265288202/1184966117941313536

a.completed_book:before {
    content: url('/static/icn/tick.png');
    margin-right: 5px;
}

but then also have the book titles line up.  Easiest/best just to put a new unnamed column in the datatables output.

robby1066:
this fixes #57 by doing the following:

1. adds a class `book-title` to all book title links, so they can all be targeted easily with css
2. adds an empty placeholder in front of all `a.book-title` links so the indention is consistent
3. shows `tick.png` in the placeholder space on all `a.book-title.completed_book` links
4. adds a selector with good-but-not-perfect browser support* that removes indentation when the list contains zero completed books. On unsupported browsers the indent will remain even if there are no completed books.

* the `:has()` selector works in every modern browser, but firefox only added support in the most recent version (Dec, '23)

### DIFF
--- a/lute/static/css/styles.css
+++ b/lute/static/css/styles.css
@@ -141,11 +141,10 @@ div#divbooktable {
 a.book-title:before {
     content: '';
     display: inline-block;
-    offset-anchor: right;
     width: 16px;
     height: 16px;
     vertical-align: baseline;
-    margin-right: 0.5rem;
+    margin-right: 8px;
 }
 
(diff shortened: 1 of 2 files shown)


jzohrab:
Looks great!

PR #76, opened 2023-12-24
Add Hindi language definition and a small sample story.

mzraly:
Adding Hindi language definition and a small sample book.


jzohrab:
Checked it out, thanks, seems to work great!  I'll shorten the sample title a bit, right now it's twice as long as the longest.  Thank you very much!

PR #77, opened 2023-12-26
Resize panes by dragging borders

webofpies:
Ability to resize the left read pane and the right word and dict panes by dragging their borders.

Some code related to the "text width" functionality has been moved from `text-options.js` to the new `resize.js` file.

double click on the borders: for the main read pane it expands the read pane to 95% of the viewport width (but doesn't save this to the local storage). This is for temporarily expanding the pane if you do not need the dictionary for some time. Double clicking again (if the pane width is at 95%) gets it back to the previous width. 
For the right pane, double click expands the dictionary pane to 95% and the rest is the same as with the read pane.


jzohrab:
Looks great!  Adding a section in the manual for styling.

PR #79, opened 2023-12-26
Follow standards for javascript "data-" attribute names

robby1066:
**This change should not alter functionality anywhere.** Simply a code cleanup to make the attributes on the SPAN tags wrapping each word use the HTML data-* attribute standards. 

### Changed attributes

| Old Attribute      | New Attribute |
| ----------- | ----------- |
| tid      | data-text-id       |
| lid   | data-lang-id        |
| paraid  |  data-paragraph-id  |
| seid  |  data-sentence-id  |
| data_text  |  data-text  |
| data_status_class |  data-status-class  |
| data_order |  data-order  |
|  data_wid | data-wid |

in places where jQuery is used, references to `$(element).attr('data_some_attribute')` were changed to `$(element).data('some-attribute')`

where vanilla Javascript was used, references to `element.getAttribute('data_some_attribute')` were changed to `element.dataset.someAttribute`


jzohrab:
Even though we’d said it was sort or make-work, it’s still a nice change. Thank you very much. I’m going to do another scan of the full code before merging it, but it looks like you’ve covered everything. Cheers!

jzohrab:
Pulled the branch and checked the code for the attributes, I think all is covered.  Thank you very much!

jzohrab:
All of the acceptance tests passed, I gone through the code, everything seems good.  Thanks @robby1066 (Battle of Hastings?).

PR #81, opened 2023-12-26
change the old German orthography from the example text to the new one

dgc08:
The German example story still uses the old German orthography rules that were in use before 1996. Considering that most German learners learn modern German, it would be better to update the story.


jzohrab:
Good call, thank you!

Issue #13, opened 2023-10-06
[Feature] Ability to delete a term from reading screen.


disfated:
It is occasionally needed to be able to remove a term while staying on the reading page.

jzohrab:
Just adding a "delete" button next to the "Save" button would suffice as a first pass:

![Image](https://github.com/jzohrab/lute-v3/assets/1637133/fa449149-6712-44bc-98db-20dee3206b87)

A few caveats:

* if reading, this should trigger a page reload so that all of the parsed tokens are rendered correctly
* if on the term form, this should redirect to the term listing
* should show a warning that this can't be undone.  In particular, if a term has children, those will be orphaned and any parent description will be lost.  Perhaps an extra warning for parents would be good.

jzohrab:
Fixed in #85 , merged into develop.

jzohrab:
Launch in 3.0.9.

PR #85, opened 2023-12-28
Fix issue #13: Ability to delete a term from reading screen

dgc08:
Added delete term button in term form, visible in both `/read` and `/term/edit`:
 Button in `/read` reloads the page
 Button in `/term/edit` redirects to `/term/index`

Removed old delete button in `/term/edit`

Check if you're fine with the layout, I've placed the button below the "save" button, in the next table row.
This is the first JS code I have ever written lol, but it works.


jzohrab:
Ah, changing the delete back to a POST though, it's really a POST action, which might change the "fetch" back to an ajax post.  All good.

PR #88, opened 2023-12-29
Read slider: slider to turn pages

webofpies:
Added a slider to turn pages. Turning pages with buttons as before still works. 
Bookmark button for page bookmarks created but hidden until implementation.
Screenshot_6

### DIFF
--- a/lute/static/css/styles.css
+++ b/lute/static/css/styles.css
@@ -70,7 +70,7 @@ p.version {
     display: flex;
     justify-content: space-between;
     align-items: center;
-    padding: 10px;
+    padding: 0.6rem 2rem 0 2rem;
 }
 
 .title {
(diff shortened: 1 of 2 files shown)


dgc08:
Two things I noticed during my tensting:
 - Styles (I use for example Dark Slate) are not working
 - When scrolling through a lot of pages, the scrolling is not that fluent

scrollbar1

webofpies:
Redesign:

- header is now more compact thanks to your feedback.
- color is the same as audio player main button color. not too muted because it didn't looked nice with dark themes and insignificant overall.
- added border to the slider menu button to separate it visually from the logo. i think looks better.
- text title is smaller for all pages. if it's bigger for the first page there's a jump on page turn.
- regarding the scrolling fluency: now the page turns upon mouse release, not in "real time". so this shouldn't be an issue
- slider is visually disabled if there's only one page
- part of the title overflowing the free space gets hidden

image
image

PR #113, opened 2024-01-04
Update broken links in Readme.md

robby1066:
I saw that the links in Readme.md were getting a 404. Updating them to the relevant pages in the wiki.

Using relative links, which you may want to change to absolute ones (e.g. https://github.com/jzohrab/lute-v3/wiki/Development) as wiki's don't come over when the repo is forked.


jzohrab:
Thanks, little things like this add up, so you've saved me some time.  Appreciated!

PR #118, opened 2024-01-06
Compact layout for mobile devices

webofpies:
Compact layout for mobile devices (<980px)

changes:
- reordered `reading_menu` in the html. and now its `position` is `fixed`


lute/templates/read/index.html
```diff
@@ -370,27 +414,27 @@ <h2>&#127881;</h2>
...
+    // disableScroll(false);
   }
 
   function toggleMenu() {
     const menu = document.getElementById("reading_menu");
-    menu.classList.toggle("open-menu");
+    menu.classList.add("open-menu");
 
-    disableScroll(true);
+    // disableScroll(true);
```
jzohrab, line 424: No disabling scroll needed anymore?  Should we just remove this code and related code instead of commenting it out, or are you unsure about this?

lute/themes/css/Apple_Books.css
```diff
@@ -1,10 +1,9 @@
 body {
-  background-color: #f8f1e2; /* beige paper */
+  --background-color: #f8f1e2;
+  background-color: var(--background-color); /* beige paper */
   color: #262523; /* dark grey font */
 }
 
- #reading_menu { background-color: #f8f1e2; }
```
jzohrab, line 6: Was this style redundant?  Wonder why it was there if yes.  Same comment applies to the other themes.

Also, did the themes need to change to support this PR, or is this additional/related work that you're cleaning up?  Minor point, just asking.

jzohrab (commented):
I haven't tried this code out yet, but the PR looks great, thank you!  Just a few minor questions in the review comments.  (and the CI passed, :tada: so these changes are likely good!)

One last question, which you might not know: is there a way to add different viewport dimensions in the ci tests?  Currently, the acceptance tests start the Lute app (to be driven by the tests) via `tests/acceptance/start_acceptance_app.py`, and that thing is called in the tasks in `/tasks.py` (with `inv accept`).  Perhaps there's a way to also run the exact same set of tests with a smaller viewport, just for fun.  Yeah, that's a strange idea of fun.  Part of me wants to see that added to this before we merge it, just for sanity/safety, but another part thinks that that is being overly cautious.

lute/templates/read/index.html
webofpies: after making the `reading_menu ` fixed not disabling scroll gives better results. menu is fixed, but you can still scroll the text and not add a scrollbar which was annoying. I am just doubting myself sometimes and keeping code commented until next update to see if weird bugs come out

lute/themes/css/Apple_Books.css
webofpies: I applied the color in the `styles.css` via the variable, so this became reduntant. by default it's just white, but changing the `--background-color` in these theme files changes the reading_menu color

webofpies:
you're right about the checks. I have no idea how those work, so unfortunanely cannot answer these :)

jzohrab:
I checked out the branch and am getting some odd behaviour on resize, e.g.:

https://github.com/jzohrab/lute-v3/assets/1637133/12d73d43-e8ff-460e-941d-190d1ab4ba3b

The video first shows a regular save, which works great.  When the screen is smaller, the form is too small at the bottom of the screen and I can't resize it, nor can I see the dictionary etc.  Any thoughts?

jzohrab (changes requested):
Notes about the resize causing problems -- not sure what is wrong here.  My system?  Concerned that this may bust Lute for some users, though for regular screen sizes it should be fine.

webofpies:
fixed the resize bug. also added "close" button and a dragging handle to make it easier to grab the pane.

Issue #93, opened 2023-12-30
Support PDF imports


jzohrab:
Lute suuports txt and epub imports, it would be nice to support pdf as well.

For epub imports, a user had to make a new simple library to get around licensing problems, so if using a PDF library, it should be a good license, no real restrictions.  The library should have some basic tests.

The changes in the Lute code to support PDF imports would likely be very similar to those for epub imports:

```
# . ---------------
./requirements.txt:openepub==0.0.6
./pyproject.toml:  "openepub>=0.0.6,<1"
# lute ---------------
lute/book/service.py:def get_epub_content(epub_file_field_data):
lute/book/service.py:    Get the content of the epub as a single string.   .... etc

lute/book/forms.py:                ["txt", "epub"],
lute/book/forms.py:                "Please upload a valid .txt or .epub file.",

lute/book/routes.py:    if ext == ".epub":
lute/book/routes.py:        return service.get_epub_content(filefielddata)

lute/templates/book/create_new.html:      <td>{{ form.textfile.label }} <i>(.txt, .epub)</i></td>

# tests ---------------
tests/acceptance/book.feature:    Scenario: I can import an epub file.
tests/acceptance/book.feature:        Given a Spanish book "Hola" from file Hola.epub
```

dgc08:
PyPDF2 seems like a good library: https://github.com/py-pdf/pypdf

jzohrab:
Is the license ok?

dgc08:
As I understand it, we just have to include their copyright notice and then we're fine.

jzohrab:
Merged into `develop` :tada:

PR #119, opened 2024-01-07
Add PDF import support for books (Issue #93)

dgc08:
Implement issue #93 

I took the PR for the epub support as template and chnaged it accordingly.

 I used `PyPDF2` as library, the license should be fine: [ https://pypi.org/project/PyPDF2/](https://pypi.org/project/PyPDF2/)


jzohrab:
Hi @dgc08 -- this failed one of the acceptance tests, can you take a look?

jzohrab (changes requested):
Failed acceptance test.

dgc08:
The acceptance test should work now. The sample PDF file contained an extra page number, which also got imported. Didn't run the test beforehand, sorry.

jzohrab:
No problem that’s what the tests are there for. Thanks for adding the test
it’s a big help for stability.

dgc08:
kinda screwed up my fork, wait a moment before merging

dgc08:
ok its fine now

dgc08:
I hope that's it. Thank you for your patience with me, Lute is the first Open Source project / larger project in general that i contribute to, pytest and that stuff is still new to me

jzohrab:
It’s a super contribution so thank you. I’ll try it out soon with a bigger
pdf to see how it works, and will review the code again.

For a first PR it’s great!  👍

jzohrab:
Hi @dgc08 -- I did an import of a large file and found a few places where the import adds some spaces, eg

> image

vs the original pdf:

> image

This happens fairly frequently.  It's almost certain that it's due to the PDF parser library ... should check.

jzohrab:
Yes, it's the library, there's a good write up by the author(s) in these links:

* https://stackoverflow.com/questions/76546531/why-does-pypdf-stuff-text-with-extra-spaces-when-extracting-text
* https://github.com/py-pdf/pypdf/issues/1507

Per the authors:

> Getting whitespaces right is notoriously hard. @pubpub-zz is the expert in that topic; I'll leave it to him to decide if we should leave this issue open. The issue is that PDF does not (necessarily) represent the words as words internally. In the worst case, it just gives the absolute position of each character in the document.

So, I think this should be fine as it is, but we should mention somewhere that PDF imports are extremely tricky.

jzohrab:
Added a flash:

> image

Issue #120, opened 2024-01-08
Add --update option to lute/main.py that calls sys.executable -m pip install -y lute3


mzraly:
When a new lute update is available, it should be possible to update with a single command.

I normally run lute using virtualenv with a wrapper script that activates the right environment.  No doubt there are other users doing something similar, adding a script that sets up the environment and passes command-line arguments to the python -m lute.main command.

At the moment, upgrading the python environment requires establishing the environment (activating a virtual env, running via docker, setting environment variables, or whatever) and THEN calling pip to upgrade.

Adding a --upgrade option to lute.main would let all these users update using the same wrapper.

Implementation should be simple:

1. Import sys and subprocess
2. Add a new boolean --upgrade option
3. Upgrade using: subprocess.run([sys.executable, "-m", "pip", "install", "--upgrade", "lute3"], check=True)

jzohrab:
I think I'd like to keep main.py strictly for starting Lute ... but hey, that's just me, and I can ignore the option I suppose!

In addition to the code change, please suggest the relevant documentation changes to the Lute manual.  https://github.com/jzohrab/lute-manual

Docker uses shouldn't update their lute, they should either use the pre-built image, or build their image from source.

mzraly:
Hmm, I hadn't thought about the docker image, and updates not being persistent.  I updated the --help output accordingly.

Would you like a pull request to update install/install.md?  Or should I just write suggested changes here?

jzohrab:
Do we even need this at all?  The page
https://jzohrab.github.io/lute-manual/usage/updating.html covers what's
needed, and any wrapper scripts are going to need to add the flag etc.

mzraly:
Fair enough, let's kill this issue and PR.

jzohrab:
Closing, thanks -- but perhaps there is something that should be done for boilerplate startup scripts for each environment.  I have my mac script, isán posted something about a windows/powershell, and I think another user had a linux script.  I'll open a separate issue for that, some people may find it useful.

PR #121, opened 2024-01-08
Add --upgrade option, addressing #120

mzraly:
Let's user run "python -m lute.main --upgrade" to upgrade to latest version using pip.

### DIFF
--- a/lute/main.py
+++ b/lute/main.py
@@ -8,10 +8,13 @@ e.g.
 python -m lute.main --port 5001
 """
 
-import os
 import argparse
-import shutil
 import logging
+import os
+import shutil
+import subprocess
+import sys
+
 from waitress import serve
 from lute.app_factory import create_app
 from lute.config.app_config import AppConfig
@@ -99,7 +102,7 @@ def _start(args):
     serve(app, host="0.0.0.0", port=args.port)
 
 
-if __name__ == "__main__":
+def _main():
     parser = argparse.ArgumentParser(description="Start lute.")
     parser.add_argument(
         "--port", type=int, default=5000, help="Port number (default: 5000)"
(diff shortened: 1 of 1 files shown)


jzohrab:
This needs some docs -- please suggest the relevant documentation changes to the Lute manual.  https://github.com/jzohrab/lute-manual

Thanks for the PR!

lute/main.py
```diff
@@ -108,12 +111,34 @@ def _start(args):
         "--config",
         help="Path to override config file.  Uses lute/config/config.yml if not set.",
     )
+    parser.add_argument(
+        "--upgrade",
+        action="store_true",
+        help="Upgrade to latest version (doesn't work with docker)",
```
jzohrab, line 117: Not exactly accurate: this shouldn't be used with Docker, the image should be rebuilt.

lute/main.py
```diff
@@ -108,12 +111,34 @@ def _start(args):
...
+    )
+
     try:
-        _start(parser.parse_args())
+        args = parser.parse_args()
+        if args.upgrade:
```
jzohrab, line 122: This command means that lute.main has different behaviours based on flags: `python -m lute.main --upgrade --port 5000` won't actually start lute, it only upgrades.  Should the `--upgrade` flag first run the update process, and then start lute?

The more I think about this, the more I feel that the "update" process should be handled outside of lute.main, and inside of any wrapper script that people use to start Lute.  For example, for my mac, I have the following:

```
#!/bin/bash

# Start venv, pull a version from pypi if specified, and start lute with this config.

source .venv/bin/activate

VERSION="$1"
if [[ -z "$VERSION" ]]; then
    echo "Using existing install."
else
    echo "Pulling version $VERSION"
    pip install --upgrade lute3==${VERSION}
fi

open http://localhost:9876
python -m lute.main --port 9876
```

jzohrab (commented):
Not sure about this change, should we just have the user wrapper scripts handle it?  They'd have to pass flags anyway to get things to run.

jzohrab:
Closing per discussion in issue -- but we should probably have a "scripts" folder or something for people to take for their environments.

PR #123, opened 2024-01-10
Rework/optimize form and table styles

webofpies:
Standardized form styles.
Optimized `styles.css` by removing unused, unneeded styles and by combining styles where possible


jzohrab:
Thank you very much, looks super 👍

PR #127, opened 2024-01-11
Z-index order fix

webofpies:
Fixes the issue introduced after mobile view update. 
image

Main issue was that the tooltip was being created under the audio player if the word was close to the bottom of the page. Now it's created over the player. 
But a better solution would be detecting the audio player as the "end of the page" and showing the tooltip above the word.


jzohrab:
All makes sense, thanks!

PR #143, opened 2024-01-17
Make the mecab_parser.py also support Unidic

fanyingfx:
To make mecab_parser support Unidic
I also make a tutorial on how to change Mecab dictionary  to Unidic : https://fanyingfx.github.io/note/Using-Mecab-with-Unidic
If user using mecab default dictionary, nothing should be different.
The Unidic works well when parse paragraphs, only get error when get reading.
This pr is making JapaneseParser can also get reading from Unidic by changing the mecab flags.

There is another problem, if using the Unidic, some Japanese test cases will fail, because different dictionaries can get different parsed results. I don't know how make the appropriate test cases for such scenarios , and change the mecab dictionary is out of Lute's scope.

### DIFF
--- a/lute/parse/mecab_parser.py
+++ b/lute/parse/mecab_parser.py
@@ -16,6 +16,7 @@ import os
 import re
 from typing import List
 from natto import MeCab
+from natto.api import MeCabError
 import jaconv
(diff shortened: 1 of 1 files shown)


fanyingfx:
Find another solution, do not need to change the code.

PR #148, opened 2024-01-20
Replace symbols with SVG

webofpies:
image


jzohrab:
Merged manually, thanks!

PR #159, opened 2024-01-26
Booktable fixes

webofpies:
Adjusted status bar for small views removing percentages below 2
Fixed book name wrapping


jzohrab:
Thank you!

PR #161, opened 2024-01-27
Consistent field styles

webofpies:
Form fields and other input types (including tagify fields and the new "words per page" field) styled to be visually consistent between one another


jzohrab:
Thanks @webofpies :-)  A comparison:

The book creation thing looks much better, thx:

> image

Re the term form:

This PR:

> image

Develop:

> image

The only changes I notice are:

* the selected field outline is narrower, which is nice
* the tags look more squished, which I don't like

webofpies:
increased padding of individual tags
image

jzohrab:
Thanks, super 👍

PR #163, opened 2024-01-28
Booktable fix

webofpies:
Fixed the issue (introduced in the PR related to refresh all stats button) where Statuses column was getting distorted if browser window was narrower.

image


jzohrab:
Thanks for the fix!  In future I'll try to remember to resize to check things out.

PR #224, opened 2024-02-09
Dictionary tabs

webofpies:
A tab for each dictionary


lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+  if (src) img.src = src;
+
+  return img;
+}
+
+function isURLExternal(dictURL) {
+  return (dictURL.charAt(0) == '*') ? 1 : 0; 
+}
+
+function getDictURLDomain(url) {
```
jzohrab, line 324: Update to handle "invalid" urls.

The tests failed because I had my `/dev_api/` url, which isn't valid :slight_smile:  In future the URLs might not be http URLs, so we can change that method to the below, and then everything passes:
```
function getDictURLDomain(url) {
  const clean_url = url.split("*").splice(-1)[0];
  try {
    const urlObj = new URL(clean_url);
    return urlObj.hostname;
  }
  catch(err) {
    // Handle test/non-http dictionaries.
    let d = clean_url.slice(0, 10);
    if (d.length < clean_url.length)
      d += '...';
    return d;
  }
}
```

(ps -  If one acceptance test fails, it could just be a bad test, but if many of them fail, it generally means something has changed in the code in a bad way)

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
+"use strict";
+
+function createDictTabs(num = 0) {
+  // TERM_DICTS.push("https://de.thefreedictionary.com/###");
```
jzohrab, line 4: Please remove dead code (don't be like me).

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+  if (num == 0 || num == null || num >= TERM_DICTS.length) {
+    sliceIndex = TERM_DICTS.length;
+    columnCount = sliceIndex;
+  } else {
+    sliceIndex = num - 1;
+    columnCount = num;
+  }
+  
+  const TABBED_DICTS = TERM_DICTS.slice(0, sliceIndex);
+  const OPTION_DICTS = TERM_DICTS.slice(sliceIndex);
```
jzohrab, line 29: It's not clear what OPTION_DICTS is, or the sliceIndex etc.

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+  // TERM_DICTS.push("*https://dict.tu-chemnitz.de/deutsch-englisch/###.html");
+  // TERM_DICTS.push("*https://www.translate.ru/%D0%BF%D0%B5%D1%80%D0%B5%D0%B2%D0%BE%D0%B4/%D0%BD%D0%B5%D0%BC%D0%B5%D1%86%D0%BA%D0%B8%D0%B9-%D0%B0%D0%BD%D0%B3%D0%BB%D0%B8%D0%B9%D1%81%D0%BA%D0%B8%D0%B9/###");
+
+  if (TERM_DICTS.length <= 0) return;
+
+  let sliceIndex;
+  let columnCount;
+
+  if (num < 0) num = 0;
+  if (num == 1) num = 2;
```
jzohrab, line 18: This is unclear.  Why 2 tabs, if only 1 tab?  Shouldn't < 0 tabs throw an error?

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+      option.classList.add("dict-select-option");
+      const origIndex = TERM_DICTS.indexOf(dict);
+      const domain = getDictURLDomain(dict);
+      const faviconURL = getFavicon(domain);
+      const faviconEl = createImg(faviconURL, "dict-btn-fav-img");
+      option.textContent = domain.split("www.").splice(-1)[0];
+      option.prepend(faviconEl);
+      option.dataset.dictId = origIndex;
+      option.dataset.dictExternal = isURLExternal(dict);
+      option.dataset.tabOpened = 1 - isURLExternal(dict);
```
jzohrab, line 97: This line is very hard to understand.  Do you mean it should open a tab, or that a tab is already open?  And isUrlExternal == 1 or 0 only?  Is this even necessary, given that the previous line says "dictExternal = isURLExternal()" ?

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+    dictTabsLayoutContainer.appendChild(selectButtonBox);
+  
+    OPTION_DICTS.forEach((dict) => {
+      const option = document.createElement("p");
+      option.classList.add("dict-select-option");
+      const origIndex = TERM_DICTS.indexOf(dict);
+      const domain = getDictURLDomain(dict);
+      const faviconURL = getFavicon(domain);
+      const faviconEl = createImg(faviconURL, "dict-btn-fav-img");
+      option.textContent = domain.split("www.").splice(-1)[0];
```
jzohrab, line 93: Breaks for non-www domains (here and similar lines elsewhere).

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+  
+  // create image button
+  const imageBtn = createTabBtn("", dictTabsContainer, -1, 0);
+  imageBtn.setAttribute("id", "dict-image-btn");
+  imageBtn.setAttribute("title", "Look up images for the term");
+  const imageFrame = createIFrame("imageframe", iFramesContainer);
+
+  dictTabButtons.set(imageBtn, imageFrame);
+  
+  // set first frame as active (for final: need to save active tab and retrieve it)
```
jzohrab, line 166: Checking on the comment "need to save active tab and retrieve it" -- we don't need to save the active tab, do we?

lute/templates/term/formframes.html
```diff
@@ -12,10 +12,64 @@
...
-  </iframe>
+  <div class="dictcontainer">
+    <div id="dicttabs">
+      <div id="dicttabslayout"></div>
+    </div>
+    <div id="dictframes"></div>
+  </div>
 </div>
 
+<!-- <script>
```
jzohrab, line 23: Is this commented out code, can it be removed?

lute/templates/term/formframes.html
```diff
@@ -12,10 +12,64 @@
...
+{% if term.language_id is none %}
+
+  <script>
+    const loadDictsBtn = document.createElement("button");
+    const languageDiv = document.getElementById("languageSel");
+    const langSelect = document.getElementById("language_id");
+    loadDictsBtn.setAttribute("title", "Load dictionaries for the new term");
+    loadDictsBtn.setAttribute("id", "load-dicts-btn");
+    loadDictsBtn.setAttribute("type", "button");
+    languageDiv.appendChild(loadDictsBtn);
```
jzohrab, line 46: Button for new terms, thx.  It wasn't clear on initial read of the code.  I wonder if this would be simpler to just add to the term form ... but maybe not, maybe appending via JS is the better way.

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+      
+      btn.dataset.dictId = optionVal;
+      btn.dataset.dictExternal = clickedOption.dataset.dictExternal;
+      btn.textContent = btnLabel;
+      btn.dataset.tabOpened = clickedOption.dataset.tabOpened;
+      btn.prepend(faviconEl);
+      const menuImgEl = createImg("", "dict-btn-list-img");
+      btn.appendChild(menuImgEl);
+      
+      if (clickedOption.dataset.dictExternal == 1) {
```
jzohrab, line 140: is dictExternal always 1 or 0?  If so, best to change it to "true" or "false"

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+  loadDictPage(dictID, iFrame);
+
+  dictContainer.style.display = "flex";
+  dictContainer.style.flexDirection = "column";
+}
+
+function addSentenceBtnEvent(dictTabButtons) {
+  const iframe = document.querySelector("iframe[name='dict0']");
+  const sentencesBtn = TERM_FORM_CONTAINER.querySelector("#term-button-container > a");
+  const tab0 = document.querySelector("button[data-dict-id='0']");
```
jzohrab, line 220: Have to handle no dicts.

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+
+  sentencesBtn.addEventListener("click", (e) => {
+    e.preventDefault();
+    const url = getSentenceURL();
+    if (!url) return;
+
+    // set to "opened" so clicking on the tab reloads dictionary
+    tab0.dataset.tabOpened = 0;
```
jzohrab, line 228: Prefer using true or false for booleans.

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+  return `http://www.google.com/s2/favicons?domain=${domain}`;
+}
+
+/**
+   * Either open a new window, or show the result in the correct frame.
+   */
+function show_lookup_page(dicturl, text, iframe) {
+  // if iframe is provided use that, else it's an external link
+
+  // const is_bing = (dicturl.indexOf('www.bing.com') != -1);
```
jzohrab, line 341: Shouldn't be commented out, but the existing dictionaries stored in the db should change.

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+    // TODO zzfuture fix: fix_language_dict_asterisk
+    // The URL shouldn not be prepended with trash
+    // (e.g. "*http://" means "open an external window", while
+    // "http://" means "this can be opened in an iframe."
+    // Instead, each dict should have an "is_external" property.
+    loadPopupDictionary(dicturl, text);
+  }
+}
+
+function loadIFrameDictionary(dicturl, text, iframe) {
```
jzohrab, line 363: 2-line method only used in one place, can be moved into where it's used.

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+    use_text = parents[0];
+
+  const raw_bing_url = 'https://www.bing.com/images/search?q=###&form=HDRSC2&first=1&tsc=ImageHoverTitle';
+  const binghash = raw_bing_url.replace('https://www.bing.com/images/search?', '');
+  const url = `/bing/search/${LANG_ID}/${encodeURIComponent(use_text)}/${encodeURIComponent(binghash)}`;
+  iframe.setAttribute("src", url);
+  return;
+}
+
+/** Parents are in the tagify-managed #parentslist input box. */
```
jzohrab, line 419: Should this function be in here?  get_parents is called from this file as well as from `lute/templates/term/_form.html`.

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+  const pdata = $('#parentslist').val();
+  if ((pdata ?? '') == '') {
+    return [];
+  }
+  const j = JSON.parse(pdata);
+  const parents = j.map(e => e.value);
+  return parents;
+};
+
+// function getDictTabSetting() {
```
jzohrab, line 434: Commented code, clutters up the PR. :-)

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,445 @@
...
+
+function createImg(src, className) {
+  const img = document.createElement("img");
+  img.classList.add(className);
+  if (src) img.src = src;
+
+  return img;
+}
+
+function isURLExternal(dictURL) {
```
jzohrab, line 320: Boolean functions should return booleans.  `if (isURLExternal(url)) { }` is easier to follow than `if (isURLExternal(url) == 0) { }`.

lute/static/js/dict-tabs.js
jzohrab: Unresolving, I just find this confusing.  num is an argument that's passed, I don't follow why it's changed to 2 if it's 1.

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,455 @@
...
+    let iFrame = null;
+    let btnLabel = domain.split("www.").splice(-1)[0];
+
+    const btn = createTabBtn(btnLabel, 
+                              dictTabsLayoutContainer, 
+                              index, 
+                              isURLExternal(dict), 
+                              faviconURL);
+
+    if (!isURLExternal(dict)) {
```
jzohrab, line 49: Do we really need a separate iframe for each dict, given that they're all displayed in the same location?  e.g. if we want to show something else in the frame (per new issue https://github.com/jzohrab/lute-v3/issues/237), will we need to create another iframe?

lute/static/js/dict-tabs.js
```diff
@@ -0,0 +1,455 @@
...
+
+function createDictListMenu(dicts) {
+  const listContainer = document.createElement("div");
+  listContainer.setAttribute("id", "dict-select-container");
+  listContainer.classList.add("dict-select-list-hide");
+  
+  dicts.forEach((dict) => {
+    const menuItem = document.createElement("p");
+    menuItem.classList.add("dict-select-option");
+    const origIndex = TERM_DICTS.indexOf(dict);
```
jzohrab, line 208: Code tip: this line is very short, and the var `origIndex` is only used once, so you could skip creating a var and just assign it later, like `menuItem.dataset.dictId = TERM_DICTS.indexOf(dict);`.  It's still very clear what's up.  Same goes for other vars throughout. :-)

lute/static/js/dict-tabs.js
jzohrab: Addressing in https://github.com/webofpies/lute-v3/pull/1

lute/static/js/dict-tabs.js
jzohrab: Thanks, it's a bool!

jzohrab:
Closing in favor of #248 , thank you!

Issue #37, opened 2023-11-28
Improve backup visibility


jzohrab:
I want to know about my backups:

- show listing of files
- show size of backups

sakolkar:
Performed a first pass on this ticket.


#### Backup drop down in the navigation
* Adjusted to have a "time since" description to quickly get a sense of duration since the last backup. Sometimes converting the date directly take a bit more effort.
* Also, the create new backup has been moved into the backups listing page
* There is a link to go to the backups listing page
![image](https://github.com/jzohrab/lute-v3/assets/9434751/8e399564-c708-436c-bf03-f8b2528aa758)


#### Backup listing page
* Sorted with most recent backup at the top of the list
* Shows manual and automated backups in the same table
* Human readable size of the backup file
* Last modified on the file. Created on is a little more nuanced and requires a different method for each OS so I didn't include that.
![image](https://github.com/jzohrab/lute-v3/assets/9434751/570b392c-4187-4e71-9a69-686beb37794e)

jzohrab:
pr merged, thanx @sakolkar , it's exactly what I wanted for myself :-)

PR #227, opened 2024-02-10
#37 - More insights into backup files.

sakolkar:
* Added human readable time since in the backup drop down in the nav
* Added a listing page to list all database backup files.


jzohrab:
Thanks @sakolkar - I read through the code and it's extremely clear what's going on.

I'm currently hacking through another big PR, so I won't merge this one right away.  Hope to do so soon-ish, but with travel etc it might take longer than expected.  Cheers!

sakolkar:
Sounds good! In the meantime, I'll add some unit tests for the new bits of code.

lute/backup/service.py
```diff
@@ -20,6 +22,54 @@ class BackupException(Exception):
...
+        4 kB
+        27 MB
+        """
+        s = self.size_bytes
+        if s >= 1e9:
+            return f"{round(s * 1e-9)} GB"
+        if s >= 1e6:
+            return f"{round(s * 1e-6)} MB"
+        if s >= 1e3:
+            return f"{round(s * 1e-3)} kB"
```
jzohrab, line 69: kB => KB to match case?  (I feel so nitpicky even asking)

lute/models/setting.py
```diff
@@ -225,3 +226,40 @@ def last_backup_display_date(self):
...
+                if multiples > 1:
+                    message += "s"
+                break
+        else:
+            message = f"{abs(delta)} seconds"
+
+        if delta >= 0:
+            message += " ago"
+        else:
+            message += " from now"
```
jzohrab, line 263: Why would time since last backup return a data in the future?  Weird date/time changes, daylight savings? :-)

If this execution path is valid, and I were a user, this would feel like a bug.  I'm not sure what the right thing to display would be if delta < 0.

lute/models/setting.py
```diff
@@ -225,3 +226,40 @@ def last_backup_display_date(self):
     def get_backup_settings():
         "Get BackupSettings."
         return BackupSettings()
+
+    @property
+    def time_since_last_backup(self):
```
jzohrab, line 231: A simple unit test for this would be good, just as a sanity check.  And perhaps also to explore the different cases of times in the future, etc ... but that kind of feels like overkill.

lute/backup/service.py
```diff
@@ -20,6 +22,54 @@ class BackupException(Exception):
...
+
+        self.filepath = filepath
+        self.name = name
+        self.is_manual = self.name.startswith("manual_")
+
+    def __lt__(self, other):
+        return self.last_modified < other.last_modified
+
+    @property
+    def last_modified(self) -> datetime:
```
jzohrab, line 46: Question for you: I haven't used return types for methods like you have here (`-> datetime`), but I like the concept.  In the past I was running into annoying circular import problems when I tried using these for custom types.  Any recommendation?  Is this information used only for type hints, or is it also picked up by pylint and other tools?

lute/templates/backup/index.html
```diff
@@ -0,0 +1,37 @@
...
+      <td>{{ backup.last_modified }}</td>
+    </tr>
+    {% endfor %}
+  </tbody>
+</table>
+
+{% endif %}
+
+<br />
+<a href="{{ url_for('backup.backup', type='manual') }}">Create new</a>
```
jzohrab, line 35: Ah you moved the "create backup" link here.  Let's also keep it in the top-level menu, I like having it readily available.

lute/templates/base.html
```diff
@@ -102,9 +102,11 @@ <h1 id="luteTitle" title="Learning Using Texts">
         <span>Backup</span>
         <ul class="sub-menu last-sub-menu">
           {% if backup_last_display_date is not none %}
-          <li>last backup: {{ backup_last_display_date }}</li>
+          <li>Last backup was {{ backup_time_since }}.</li>
+          <li>{{ backup_last_display_date | format }}</li>
           {% endif %}
-          <li><a href="/backup/backup?type=manual">Create backup</a></li>
```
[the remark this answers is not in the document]
jzohrab: The `<hr>` belongs in the if/endif block above it, prevents empty line.

lute/templates/backup/index.html
```diff
@@ -0,0 +1,37 @@
...
+  <tbody>
+    {% for backup in backups %}
+    <tr>
+      <td>{{ backup.name }}</td>
+      <td>{{ backup.size }}</td>
+      <td>{{ backup.last_modified }}</td>
+    </tr>
+    {% endfor %}
+  </tbody>
+</table>
```
jzohrab, line 30: We can style the table a bit differently -- remove some of the lastmodified milliseconds, ensure the size column doesn't wrap etc -- but that's for later.  This is super!

> image

lute/templates/base.html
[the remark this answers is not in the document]
jzohrab: Looks great!

> image

lute/backup/routes.py
```diff
@@ -7,12 +7,24 @@
...
 
 
+@bp.route("/index")
+def index():
+    """
+    List all backups.
+    """
+    settings = BackupSettings.get_backup_settings()
+    backups = list_backups(settings.backup_dir)
+    backups.sort(reverse=True)
```
jzohrab, line 23: For longtime users like me, the list might get long.  The automatic backup list is pruned regularly, but the manual ones are not.  Limiting the list should be for a later iteration.

jzohrab (dismissed):
Hey @sakolkar -- super stuff, thank you very much!  I've requested a few changes, and also had some questions that don't need code changes.

Let me know how your time is, it may be just as fast for both of us for me to pull in your branch and do the changes, rather than have you make time to do the simple things.  You've already done all of the hard work.

Thank you again!

lute/templates/base.html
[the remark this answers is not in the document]
sakolkar: I can make these tweaks!

lute/models/setting.py
sakolkar: On WSL whenever my laptop goes to sleep it falls back in time and then I run a time-sync command. But if two commands run at the same time, then it throws my time into the future until I run a time-sync again.


So... its a contrived way to get into future time. One would have to run a backup in that duration and then correct time and then go see that it says "from now". DST and timezones wont affect it, however, since `time.time()` is Unix time I believe.


We could probably put a more safe value so users aren't confused by something saying it happened in the future or thinking that thats when the backup is scheduled for.


How about no human readable value if its in the future? It could still have the original timestamp there though.

lute/backup/service.py
sakolkar: Yes, the typing system really helps to jump to definition and see references. When you run into cyclical imports, the typing allows you to specify it as a string.

eg. `-> datetime.datetime` is equivalent to `-> "datetime.datetime"`

The first would require the import but the second has it differed.

sakolkar: It is also picked up by the linters and in VSCode it will underline blue when you write code that doesnt respect the types.

lute/templates/backup/index.html
sakolkar: Since we're touching this code, I can do this adjustment now. Also that's fairly easy to implement!

lute/backup/routes.py
sakolkar: Yep, I was thinking maybe there would be a delete option given for manual backups in the future.

lute/models/setting.py
jzohrab: Sure, no human-readable thing, to prevent my computer from time travelling :-)

lute/backup/service.py
jzohrab: Caps looks better I think, given the other caps.

lute/backup/service.py
jzohrab: Thanks for the info.  I haven't really set up a good editing environment yet, just emacs and treemacs, but it's good enough for me.  But the type hinting is good.

lute/models/setting.py
sakolkar: Updated so that the human readable portion is hidden if the time is in the future.

image

jzohrab:
@sakolkar - don't worry about that failed acceptance test in the CI pipeline, it's not related to your changes.  It fails intermittently and I'm not sure why.

Let me know when you think it's good to go, and I'll check it out again locally.  Thanks for making the updates :-)

sakolkar:
Hey, ok yea. I think the current failures are with some unit tests I've been adding. Probably this afternoon, I can wrap those up and have you review once more then.

jzohrab:
If the failure has to do with acceptance testing, then 99.9% sure you can ignore it, just do a quick check.

Thanks, this is going to be a nice add that will simplify a few other things I've wanted to do. 👍

sakolkar:
@jzohrab - the unit tests are all in and other items you mentioned have been tweaked as well. Its now just failing on the acceptance tests like you mentioned.

jzohrab:
Thanks very much, I'll try this one out again.  I (hope I) have fixed that failing acceptance test in dev.  Thank you!

jzohrab:
Looks super!  Very nice having the "last backup" data present, and the table listing is great.

jzohrab:
Looks great, code is very clear, tests present and passing -- what more could one ask for?  👍  Thank you very much!

sakolkar:
Yay! Thank you! 😊

PR #248, opened 2024-02-12
Dict tabs revisions

jzohrab:
From #224 , seeing if I can add anything useful to the code.  This code is tough!  Draft PR for reviewing.

Current state -- all of the UI magic was from @webofpies 👍 

> image

* I added extra dummy URLs to the term lookup js just to force more elements
* the last button is a drop down, and obviously the drop down is in the wrong place.   b/c I suck at CSS.

---

The biggest change of this PR from PR 224 so far: rather than have a separate "data structure" for buttons in tabs and buttons in the list, the current code just creates the same buttons for each, and then partitions them.  This currently causes a problem when trying to find the iframe, but that should be easy to fix.

The current things I don't like about the code as it stands:

**Tabs vs buttons:** The use of tabs for some buttons, and a drop-down list for others, feels like a premature layout optimization.  It creates two different code setup paths, and different kinds of event handlers.  I combined them because they're both buttons, but still the need for extra events etc remains, along with the cloning/list management for the buttons in the list.  It's very nice, but at this point I don't want very nice, I want easy and clear. 🙂

**Repeated code:** There are some things that don't feel right.  For example, the calls

```
        dictTabButtons = createDictTabs(5);
        loadDictionaries(dictTabButtons);
        addSentenceBtnEvent(dictTabButtons);
```

are more-or-less repeated in 3 places, and they expose internal things (that is, the dictTabButtons) to other pages, so changing that structure would break those other pages. 

**Misleading vars:** The current code has a bunch of data structures to manage things, and they're sort of hard to grok.   For example, the `dictTabButtons` sounds like an array; but it's more like a `dict_tab_buttons_to_iframe_map` ... but not every button has an iframe, and this should all be managed in the buttons instead as objects (my preference, I feel it will be a lot clearer).

**Button clicks are handled globally:** Combining all buttons (term lookups, sentence lookups, image lookups) -- discussed earlier.  It would be a lot clearer if each button had a "onclick" handler just assigned, and each "on click" handler clearly defined.  Right now, many things appear to be delegated to a central handler which then figures out the source that was clicked.

## Suggestion: use "button objects"

To me, the buttons are a good place where to use javascript objects, because then you could just have everything in the button itself:  e.g. you don't have to call `if b.dataset.isExternal { iframe = frameMap.get(b) ...}`, you could just say: `f = b.iframe`: the button constructor would have all of that already set for you.  It's good information hiding and encapsulation, it's clear and easy to use.

```
const all_buttons = make_buttons(...);
all_frames = all_buttons.select(b => b.iframe).map(b => b.iframe);
... etc etc
// or better
button.load_frame();
// can keep track of if it's loaded or not, and it knows its own frame;
// also knows to open a new window, though that might start to get to
// be overcomplicated.
```

## About the merge conflict

This branch conflicts with develop due to changes introduce in the dictionary management.   I did a test merge of `develop` into the dict_tabs branch (the one before I started doing my hacking). `lute/templates/term/_form.html` had a conflict. but it was only because I changed a single line in that file, uncommenting and changing the height of the popup:

```
-      //'width=800, height=400, scrollbars=yes, ... etc
+      'width=800, height=600, scrollbars=yes, ... etc
```

There was a small styles conflict, that was no problem though.

### DIFF
--- a/lute/static/css/styles.css
+++ b/lute/static/css/styles.css
@@ -1402,6 +1402,7 @@ input[type="checkbox"][disabled] + label {
 
     position: absolute;
     background-color: aliceblue;
+    min-width: 100%;
     top: 100%;
     left: 0;
     border-bottom-left-radius: 3px;
@@ -1432,6 +1433,9 @@ input[type="checkbox"][disabled] + label {
     align-items: center;
     gap: 0.3rem;
     cursor: pointer;
+    border: none;
+    background: none;
+    width: 100%;
 }
 
 .dict-menu-item:hover {
--- a/lute/static/css/styles.css
+++ b/lute/static/css/styles.css
@@ -1121,11 +1121,6 @@ div.help-text {
     width: 100%;
 }
 
-.term-element-box {
-    display: flex;
-    gap: 0.5rem;
-}
-
 #term #languageSel {
     display: flex;
     justify-content: space-between;
@@ -1165,12 +1160,10 @@ input[type="checkbox"][disabled] + label {
 
 #term-button-container {
     display: flex;
-    justify-content: space-between;
+    justify-content: flex-end;
     align-items: center;
-}
-
-#term-button-container .term-element-box {
     gap: 0.2rem;
+    margin-top: 0.6rem;
 }
 
 .zoomableTermImage {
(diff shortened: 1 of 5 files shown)


jzohrab:
Moved to #264 .

PR #249, opened 2024-02-12
Headings for dictionary form

webofpies:
image


jzohrab:
Not adding headings, the other styling worked great, thank you!

PR #259, opened 2024-02-15
DictTab class

webofpies:
Made a DictTab class to contain all tab related info
Separated sentences and image lookup from standard tabs
Fixed menu list

### DIFF
--- a/lute/static/js/dict-tabs.js
+++ b/lute/static/js/dict-tabs.js
@@ -1,22 +1,67 @@
 "use strict";
 
+let dictTabs = [];
 
-function createTabBtn(label, dictID, external, faviconURL=null) {
-  const btn = document.createElement("button");
-  btn.classList.add("dict-btn");
-  if (label != "") {
-    btn.textContent = label;
-    btn.setAttribute("title", label);
+class DictTab {
+  constructor(dict, frameName, frameOnly=false) {
+    const dictInfo = dict ? getDictInfo(dict) : null;
+
+    this.dictID = dictInfo?.id;
+
+    this.frame = this.createIFrame(frameName);
+    
+    if (!frameOnly) {
+      this.btn = document.createElement("button");
+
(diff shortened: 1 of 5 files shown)


lute/static/js/dict-tabs.js
```diff
@@ -1,22 +1,61 @@
...
+    this.dictID = dictInfo ? dictInfo.id : null;
+
+    this.frame = this.createIFrame(frameName);
+    this.btn = document.createElement("button");
+    this.btn.classList.add("dict-btn");
+
+    if (this.dictID != null) {
+      this.label = dictInfo.label;
+      this.isExternal = dictInfo.isExternal;
+      this.btn.dataset.dictId = this.dictID;
```
jzohrab, line 17: With the class in place, it may be possible to get rid of the `dataset` stuff, and just have the information stored in the DictTab class.  This applies to the button dataset as well as the frame, eg instead of `dictTabs.forEach(tab => tab.frame.dataset.contentLoaded = "false");` you might be able to say something like `dictTabs.forEach(tab => tab.reset());` or similar.

lute/static/js/dict-tabs.js
```diff
@@ -1,22 +1,61 @@
...
-  btn.classList.add("dict-btn");
-  if (label != "") {
-    btn.textContent = label;
-    btn.setAttribute("title", label);
+  clickCallback() {
+    if (this.isExternal) {
+      load_dict_popup(this.dictID);
+      return;
+    }
+    if (this.frame.dataset.contentLoaded == "false") {
```
jzohrab, line 41: Following the dataset comment, this would become `if (!this.is_loaded())` or similar.

jzohrab (approved):
My little tweaks are just tweaks, we can do that on the main branch.   I'd say this is good to go.

jzohrab:
Thanks!

Issue #238, opened 2024-02-11
Add a 'Focus mode'


shetozz:
**Is your feature request related to a problem? Please describe.**

The right-side panel is kinda useless after you make the entries to the new terms.

**Describe the solution you'd like**

There should be a 'Focus mode' where the user can disable the right-side panel, have the text centered with some amount of padding on both sides, and the option to hide/show highlight.

**Describe alternatives you've considered**

Moving the side panel all the way to the right, but there's no option for padding now, so it makes text too wide and tedious to read.

M-Biggles:
I had thought: it would be nice if we could "dock" the right-side panel by moving the divider all the way to the side. It won't open unless you drag it back open (or maybe a little tab?). Doesn't need to load anything but the tooltips when it's closed over. Possible?

jzohrab:
In progress, pretty much done in #262 .  @M-Biggles your suggestion is good but we've gone with a "toggle" button for now.  ... Let's see how this new design idea works out.

M-Biggles:
> In progress, pretty much done in #262 . @M-Biggles your suggestion is good but we've gone with a "toggle" button for now. ... Let's see how this new design idea works out.

Sounds great! Looking forward to it :)

jzohrab:
Launched in 3.2.1.

PR #262, opened 2024-02-16
Focus mode

webofpies:
Focus mode (issue #238)

Padding of the text can be set by the "content width" buttons in the side menu.

image


lute/static/css/styles.css
```diff
@@ -987,21 +1041,19 @@ textarea {
 }
 
 /* focus on fields makes border same as tags */
-input:focus,
-textarea:focus,
-select:focus {
```
jzohrab, line 992: select:focus no longer needed?

lute/static/css/styles.css
```diff
@@ -987,21 +1041,19 @@ textarea {
...
+form input:focus,
+form textarea:focus {
     border: 1px solid var(--tagify__tag-bg) !important;
     outline: none !important;
 }
 
-input[type="checkbox"]:focus,
-input[type="file"]:focus,
-button:focus {
```
jzohrab, line 999: button:focus no longer needed?

jzohrab:
With all of the hacking for dict tabs recently merged, there are now some conflicts.  Might need some checking after resolving and merging, to ensure that the dict js doesn't throw with these changes.  Otherwise, looks super.

jzohrab (changes requested):
Just need the conflicts resolved.

lute/static/css/styles.css
webofpies: fixed

lute/static/css/styles.css
webofpies: focus border on buttons didn't look that nice. removed
also there are some buttons we don't the border or outline for. easier to remove here than add exceptions

jzohrab (approved):
LGTM 🚀

Issue #267, opened 2024-02-18
Remove lute/templates/backup/settings.html


jzohrab:
I'm _pretty sure_ this file is no longer used.  I believe it was created before the home > settings > settings was created.

Should be quick, just needs a check.

Leifen09:
I would like to work on this issue and look into it, could you please assign it to me? @jzohrab

jzohrab:
Thank you @Leifen09 :-)

PR #270, opened 2024-02-20
#267 remove backup/settings.html

EikeMenzel:
I've reviewed the usage of lute/templates/backup/settings.html and confirmed it's not utilized anywhere in our project. Following the action items outlined in Issue #267, I've removed this file to maintain code cleanliness and efficiency.

Regarding [lute/templates/settings/form.html](https://github.com/jzohrab/lute-v3/blob/master/lute/templates/settings/form.html), do we consider this file critical enough to warrant a backup? I think our old settings.html was replaced by it.

Please share your thoughts or approval on this matter so we can proceed accordingly.


jzohrab:
Hello @EikeMenzel - thank you for looking into this :-)  .

Re "do we consider this file critical enough to warrant a backup? " - can you clarify?  It's committed to git, so it's never really "lost".  Cheers!

jzohrab (approved):
LGTM, I searched as well.  Just awaiting response re "backup the file."  Thank you!

EikeMenzel:
All good, I think I just got generally confused, why this folder exists at all. I agree with you, keeping a copy shouldn't be necessary.  

From my side, you can merge it. o7

jzohrab:
Thank you very much!

PR #282, opened 2024-02-22
Fix second embedded Czech dictionary not working

imamcr:
Changed to popup.


jzohrab:
Thank you very much!

jzohrab:
Launched in 3.2.3.

PR #284, opened 2024-02-22
Fix toggle highlights classes are not removed when mouseout

imamcr:
In toggle highlight mode, cursor does not hover over the word but it still highlights it:

image

Fixed:

image


jzohrab:
Right, I don't know why I didn't do that :-)  Small but nice tweak, thank you.

jzohrab:
Launched in 3.2.3.

PR #285, opened 2024-02-22
Add LWT and LingQ themes

imamcr:
LWT theme:

image

LingQ theme:

image

### DIFF
--- a/lute/themes/css/LWT.css
+++ b/lute/themes/css/LWT.css
@@ -1,0 +1,102 @@
+body {
+  --status-0-color: #ADDFFF;
+  --status-1-color: #F5B8A9;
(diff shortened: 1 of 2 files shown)


webofpies:
I think a lot of lines are reduntant in the files, for example:

```
span.status4 {
  background-color: var(--status-4-color);
}
```

this exact rule (and many similar ones) already exists in styles.css, so there shouldn't be a need to rewrite it. redefining the vars in body{} should suffice.

jzohrab:
Those look sharp, @imamcr , thanks!   It looks like you've addressed the comments from @webofpies but I'll wait for the final 👍 from you before merging.

PR #286, opened 2024-02-24
Touch drag fix

webofpies:
Dragging to resize the panes now works correctly on touch devices
Also added drag handle to the dictionary pane

image

### DIFF
--- a/lute/static/css/styles.css
+++ b/lute/static/css/styles.css
@@ -1827,6 +1827,8 @@ input[type="checkbox"][disabled] + label {
         bottom: 0;
         left: 0;
         opacity: 0;
+        /* for touch drag resize to work */
+        touch-action: none;
     }
 
     #read_pane_right::after {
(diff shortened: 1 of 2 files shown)


jzohrab (commented):
On initial switch to this branch I had trouble with it not working due to javascript caching (presumably) -- the handles wouldn't drag, they'd just shift some screen content.  Of course, I can no longer reproduce this to demonstrate.

I had a caching problem with lute.js before so I came up with a silly hack in `app_factory.py` to prevent caching of that particular file.  It's a route that sets a never-cache header, used like this:

```
# in lute/templates/base.html

<script type="text/javascript" src="/static/js/never_cache/lute.js" charset="utf-8"></script>
```

We _could_ use that same thing in read/index.html, which currently has 

```
lute/templates/read/index.html:<script type="text/javascript" src="{{ url_for('static', filename='js/resize.js') }}" charset="utf-8" defer></script>
```

but honestly, I don't know, I'm really not sure why I had my initial problem with the switch to this code.  Perhaps it had something to do with the file create/modify times???

jzohrab:
Behaviour-wise, this looks great, once the above cache problem stopped.

jzohrab:
Started Firefox, which I had used a long time ago (I think).  With this branch, the initial drag doesn't work -- again caching??

> image

Looking into it in FF.

jzohrab:
Didn't find anything useful, and a few page refreshes fixed it.  Well, LGTM I think.  :-)

jzohrab:
Waiting to see if we get any other feedback in the discord post.

jzohrab:
Actually, forget waiting, full speed ahead.  It'll go to beta anyway.

jzohrab:
Launched in 3.2.3.

Issue #289, opened 2024-02-26
Reduce dict lookup global state


jzohrab:
Currently, the dict-tab code uses some confusing global state, we may be able to clarify some of that.

e.g. the lines 

```
lute/templates/term/_form.html:  let LANG_ID;
lute/templates/read/index.html:  const LANG_ID = {{ lang_id | safe }};
```

are both declaring a global LANG_ID variable, which term-dicts happens to use.

There are global variables and the associations aren't clear.  The globals could be explicitly moved into the dict-tabs so that their association with the code is much clearer.

Vars to find:

* TERM_FORM_CONTAINER
* TERM_DICTS
* LANG_ID

More detail:

```
(.venv) MacBook-Pro:lute-v3 jeff$ inv search TERM_FORM_CONTAINER
Searching for TERM_FORM_CONTAINER :

# . ---------------
# lute ---------------
lute/static/js/dict-tabs.js:      const txt = TERM_FORM_CONTAINER.querySelector("#text").value;
lute/static/js/dict-tabs.js:      const pdata = TERM_FORM_CONTAINER.querySelector("#parentslist").value
lute/static/js/dict-tabs.js:      const text = TERM_FORM_CONTAINER.querySelector("#text").value;
lute/static/js/dict-tabs.js:    const term = TERM_FORM_CONTAINER.querySelector("#text").value;
lute/templates/term/_form.html:  const TERM_FORM_CONTAINER = document.getElementById("term-form-container");
lute/templates/read/index.html:  let TERM_FORM_CONTAINER;
lute/templates/read/index.html:        TERM_FORM_CONTAINER = wordFrame.contentDocument.getElementById("term-form-container");
# tests ---------------
# utils ---------------
# .github ---------------
(.venv) MacBook-Pro:lute-v3 jeff$ inv search TERM_DICTS
Searching for TERM_DICTS :

# . ---------------
# lute ---------------
lute/static/js/dict-tabs.js:    this.dictID = TERM_DICTS.indexOf(dictURL);
lute/static/js/dict-tabs.js:    const dicturl = TERM_DICTS[this.dictID];
lute/static/js/dict-tabs.js:  if (TERM_DICTS.length <= 0) return;
lute/static/js/dict-tabs.js:  // TERM_DICTS.push(...dev_hack_add_dicts);
lute/static/js/dict-tabs.js:  if (tab_count == (TERM_DICTS.length - 1)) {
[...]

PR #290, opened 2024-02-26
Issue 289 clarify js globals

jzohrab:
Addresses #289.

Add `LookupButton` static vars:

```
  static TERM_FORM_CONTAINER = null;
  static TERM_DICTS = null;
  static LANG_ID = null;
```

The existing code creates global vars in a few places (either with let or const), and the LookupButton class uses these globals.  Declaring them static clarifies the intent of the variables, removes any ambiguity, and moves declaration into one place.

Many changed lines just add the LookupButton name to the var, eg `TERM_DICTS` becomes `LookupButton.TERM_DICTS`.

### DIFF
--- a/lute/static/js/dict-tabs.js
+++ b/lute/static/js/dict-tabs.js
@@ -6,6 +6,12 @@
  */
 class LookupButton {
 
+  /** State required by the buttons. */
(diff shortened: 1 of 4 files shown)


jzohrab:
Launched in 3.2.4.

PR #294, opened 2024-02-28
Word count indicator for "Create new book"

webofpies:
image

- shows number of all words and updates as you type
- shows number of selected words
- shows number of words up to the cursor

### DIFF
--- a/lute/templates/book/create_new.html
+++ b/lute/templates/book/create_new.html
@@ -30,7 +30,7 @@
     <tr>
       <td id="word-count-container">
         {{ form.text.label }}
-        <p id=word-count title="Number of: words up to cursor / all words"></p>
+        <p id=word-count title="words up to cursor(or selected) / all words">0 / 0</p>
       </td>
       <td>{{ form.text(class="form-largetextarea") }}</td>
     </tr>
@@ -100,25 +100,18 @@
     $("#language_id").change(set_field_directions);
 
     const textarea = document.getElementById("text");
-    const countDisplay = document.getElementById("word-count");
+    const wordCountEl = document.getElementById("word-count");
+
     textarea.addEventListener("input", getWordCountInfo);
     textarea.addEventListener("click", getWordCountInfo);
     textarea.addEventListener("selectionchange", getWordCountInfo);
 
-    // textarea.addEventListener("keydown", (e) => {
-    //   console.log(e.keyCode)
-    // })
-
     function getWordCountInfo() {
+      const selectionStart = (this.selectionEnd - this.selectionStart) > 0 ? this.selectionStart : 0;
       const allWordsArray = this.value.split(/\n| /).filter((item) => item);
-      const stringToCursor = this.value.slice(0, this.selectionEnd);
+      const stringToCursor = this.value.slice(selectionStart, this.selectionEnd);
       const wordsToCursorArray = stringToCursor.split(/\n| /).filter((item) => item);
-      // allWordCount = allWordsArray.length;
-      countDisplay.textContent = `${wordsToCursorArray.length} / ${allWordsArray.length}`;
-      // console.log(this.value.split(/\n| /).filter((item) => item))
-
-      // console.log(`${cursorWordCount} / ${allWordCount}`);
-      // console.log("selend", this.selectionEnd)
+      wordCountEl.textContent = `${wordsToCursorArray.length} / ${allWordsArray.length}`;
     }
   });
 </script>
--- a/lute/templates/book/create_new.html
+++ b/lute/templates/book/create_new.html
@@ -28,7 +28,10 @@
     </tr>
 
     <tr>
-      <td>{{ form.text.label }}</td>
+      <td id="word-count-container">
+        {{ form.text.label }}
+        <p id=word-count title="Number of: words up to cursor / all words"></p>
+      </td>
       <td>{{ form.text(class="form-largetextarea") }}</td>
     </tr>
 
@@ -95,6 +98,28 @@
 
     set_field_directions();
     $("#language_id").change(set_field_directions);
+
+    const textarea = document.getElementById("text");
+    const countDisplay = document.getElementById("word-count");
+    textarea.addEventListener("input", getWordCountInfo);
+    textarea.addEventListener("click", getWordCountInfo);
+    textarea.addEventListener("selectionchange", getWordCountInfo);
+
+    // textarea.addEventListener("keydown", (e) => {
+    //   console.log(e.keyCode)
+    // })
+
+    function getWordCountInfo() {
+      const allWordsArray = this.value.split(/\n| /).filter((item) => item);
+      const stringToCursor = this.value.slice(0, this.selectionEnd);
+      const wordsToCursorArray = stringToCursor.split(/\n| /).filter((item) => item);
+      // allWordCount = allWordsArray.length;
+      countDisplay.textContent = `${wordsToCursorArray.length} / ${allWordsArray.length}`;
+      // console.log(this.value.split(/\n| /).filter((item) => item))
+
+      // console.log(`${cursorWordCount} / ${allWordCount}`);
+      // console.log("selend", this.selectionEnd)
+    }
   });
 </script>
 
--- a/lute/templates/book/create_new.html
+++ b/lute/templates/book/create_new.html
@@ -107,11 +107,19 @@
     textarea.addEventListener("selectionchange", getWordCountInfo);
 
     function getWordCountInfo() {
-      const selectionStart = (this.selectionEnd - this.selectionStart) > 0 ? this.selectionStart : 0;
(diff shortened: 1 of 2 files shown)


M-Biggles:
Oh, @webofpies ! I swear I was dreaming of this! Thank you! :)

lute/templates/book/create_new.html
```diff
@@ -95,6 +98,29 @@
...
+
+    function getWordCountInfo() {
+      let allWordsArray = [];
+      let wordsToCursorArray = [];
+
+      try {
+        allWordsArray = this.value.match(/\p{L}+/gu).filter((item) => item);
+        const selectionStart = (this.selectionEnd - this.selectionStart) > 0 ? this.selectionStart : 0;
+        const stringToCursor = this.value.slice(selectionStart, this.selectionEnd);
+        wordsToCursorArray = stringToCursor.match(/\p{L}+/gu).filter((item) => item);
```
jzohrab, line 117: I don't think this will work with character-based languages or with Japanese.  This would require a server-side check, wouldn't it?

jzohrab (changes requested):

jzohrab:
I'm not clear what this adds -- users will be importing or copying/pasting text (I imagine), so what does the counter do for them?

For this to work with other languages like Japanese etc, this would need to go back to the server with an ajax call for actualy parsing.  It might get too complicated for the value it brings.

webofpies:
yes. it probably doesn't work correctly with those languages. but other than those languages, this helps me easily partition pages knowing beforehand if part of book will wholly be on a page or divided. `max words per page` number doesn't tell you much if you don't know how many words you have in the book. say you have 275 words. knowing this beforehand and and increasing the max count would be better. additionally, number of words in selection helps me cleanly separate parts of book to separate pages.  tbh, i think if current "---" break worked as page break instead of chapter break then that would have solved all of these issues. i saw that i required some indicator to tell me word count, even if it's approximate. so, thought might be useful for others too. now that you mention other languages, i can see how it looks half baked and maybe requires more work to be complete. but i can close the pr, no problem

jzohrab:
> tbh, i think if current "---" break worked as page break instead of chapter break then that would have solved all of these issues.

Isn’t that what it is?  I’ve only thought of it as a page break, people can put these breaks wherever they want.

webofpies:
maybe if there was a "ignore max word" checkbox. dashes could work as both chapter and page break

jzohrab:
> @jzohrab it IS a page break if `max words per page` number is greater than number of words between two dash breaks.

(This is probably faster to sort out on discord than here, ping me to explain what I've missed so we can work it out.)

Hey there, not meaning to be obtuse but I still don't see how I'm misunderstanding: "---" means "end the current page right at this point, and start a new one, no matter what", i.e. it's a manual page break.  It's exactly like in a Word document when you add a manual page break to force a break at a certain point; the entire doc is still paginated, taking that into account.

Just like Word, Lute adds other page breaks as required.   If `max words per page` is 5, and I give the following text:

```
hello.
---
We are going to break this up.  It will make some more pages. 
---
Thanks webofpies.
```

Then it gives 4 pages, `hello.`, `We are going to break this up.`, `It will make some more pages.`, `Thanks webofpies`.

I don't see how it could be done any differently than how it's done currently.

It feels like we have different ideas about what "page break" means -- when you say "this helps me easily partition pages knowing beforehand if part of book will wholly be on a page or divided": it sounds like you're trying to use page breaks to actually fully define a "section", i.e. you want all content between two "---" markers to be always fully included on the same page, and the book should _only_ be broken into pages using those markers.  If that's correct, then unfortunately that won't work for very long texts.  e.g. if I have a 20-page article, and for my own reasons I want each section to begin on a new page, I'd add the manual page breaks before each section start.  That shouldn't result in massive sections, those should still be broken up.

For your case of "sections", we'd need to come up with a different indicator for something like that, it's no longer a "page break", or add an "ignore max word" thing, but that could break some imports if people click it for long texts.  Regardless, I think that's a different idea than what I originally intended with the "---" manual page break marker.

webofpies:
will close this pr to explore some other options for word count or page/section break

M-Biggles:
@webofpies Would it be possible to have a counter appear at every manual break and then have a couple counters placed somewhere (floating at top? at the bottom?) for total number of words and total number of pages. If the user exceeds the page's manually set word limit, the counters could go red.

webofpies:
> @webofpies Would it be possible to have a counter appear at every manual break and then have a couple counters placed somewhere (floating at top? at the bottom?) for total number of words and total number of pages. If the user exceeds the page's manually set word limit, the counters could go red.

my pr doesn't address chinese, japanese and similar languages. so, counting words for those needs to be handled in backend, and that's additional work that needs to be explored. for now you can set the word limit to max (it's 1500 i think) and use dashes to divide and it will work as page break

M-Biggles:
> > @webofpies Would it be possible to have a counter appear at every manual break and then have a couple counters placed somewhere (floating at top? at the bottom?) for total number of words and total number of pages. If the user exceeds the page's manually set word limit, the counters could go red.
> 
> my pr doesn't address chinese, japanese and similar languages. so, counting words for those needs to be handled in backend, and that's additional work that needs to be explored. for now you can set the word limit to max (it's 1500 i think) and use dashes to divide and it will work as page break

Perhaps it would be best to attempt it with Space Delimited languages to begin with, and hook the system into the more complicated parsers for languages like those you've mentioned, at a later stage. Those languages are the ones we've got to leave notes about in our support project because they simply need another layer of tech to work best.

jzohrab:
I'd prefer not to add something that's known to be broken in some cases, just because it creates work at some point :-)

A few questions, because I don't understand this request yet.  Hoping you can fill me in so I understand what you're going for.

* I can't follow what the need is for word counts and non-breaking pages within the context of Lute as a tool for learning by reading.  Why does it matter if a page is broken up?
* re "[determining] total number of words and total number of pages" while creating the book: that will only be known by doing a full parse.  Why do you need this when you're initially creating a book?

M-Biggles:
> I'd prefer not to add something that's known to be broken in some cases, just because it creates work at some point :-)
> 
> A few questions, because I don't understand this request yet. Hoping you can fill me in so I understand what you're going for.
> 
>     * I can't follow what the need is for word counts and non-breaking pages within the context of Lute as a tool for learning by reading.  Why does it matter if a page is broken up?
> 
>     * re "[determining] total number of words and total number of pages" while creating the book: that will only be known by doing a full parse.  Why do you need this when you're initially creating a book?

I can give my own arguments for it, but I can't speak for webofpies.

1. Those languages are currently "broken" anyway, so they're going to involve more work no matter what. At least this could be made to work for the languages which do currently work in Lute - with a notice somewhere about the state of those langs, including the fact that word-counting at text-creation is currently not possible. Holding back for those which don't work, would prevent a lot of progress being made.

2. We don't currently know how many words are in the pages we're creating. It just splits up the text according to the word limit (awkwardly at times - but that's another issue). There's no marker or anything so we're just doing it hoping it looks nice when it comes to reading. There are all sorts of reasons why people want to have a bit more info about and control over how many words each page is going to have, but whatever those reasons maybe be - and I could give a few, I think that it's sensible to have a counter for a limit that's set elsewhere.

I imagine creating a new book, pasting the text into the form, and Lute automatically breaking up the text into pages by adding '---' line-breaks - according to the limit set in Words per page (I'll refer to the default of 250). Each line-break would have its counter with 250/250. Perhaps, this could be seen as a soft-limit where Lute will split texts up into 250-word segments, but where I could manually place the line breaks in another, better spot. When pages exceed the 250-word limit, Lute could ask the user 'Page(s) X has (have) exceeded the words-per-page limit. Do you wish to proceed?' If any exceed the 1,500 limit, it would give an error and ask the user to split the text further.

jzohrab:
> 1. Those languages are currently "broken" anyway,

:'-)  fair enough!

> I imagine creating a new book, pasting the text into the form

Wouldn't you upload a text file instead?

> I imagine creating a new book, ... [lots of fancy things here :-)] ...

What you're asking for feels like an application in itself, could get very touchy, and I think perhaps belongs in a different spot, rather than complicating the Lute input form.  It's outside of the scope of that input form, which is just a dumb data input, and is getting into some kind of layout manager/editor.

Boy this is a lot of work/discussion for something that was supposed to be a very small quick add. :-P

> It just splits up the text according to the word limit (awkwardly at times - but that's another issue).

yeah, this was actually a challenging thing to design.  fyi: What Lute does is break things up by full sentences, because I wanted/needed the full sentence to be present on any page, for future context lookups for terms.  Then Lute accumulates as many sentences as it can, staying within the max tokens count.  When that limit is reached, Lute says that the current set of sentences forms the page.  I know it's awkward sometimes, but it became one of those things that, in the end, just didn't matter for me, as I read tons of content.

And with that, I'll back out of this conversation!  :wave:

Issue #16, opened 2023-10-24
[Classical Chinese] preload default Dictionary 1 - clicking a word redirects to dictionary 1 entry page


farahpharos:
**Description**
When using https://ctext.org/dictionary.pl?if=en&char=### as a dictionary 1, clicking a word in the text 

**To Reproduce**

Steps to reproduce the behavior, e.g.:

1. Create new language and use Classical Chinese preload settings
2. Import a chinese language text and click on any highlighted word
3. Page will be redirected to the ctext.org definition page

This issue does not come up with other dictionaries such as:
https://www.archchinese.com/chinese_english_dictionary.html?find=###

**Extra software info, if not already included in the Description:**

- OS: Both Mac and Windows
- setup: docker 
- Version: 2.1.3 (@latest version)

jzohrab:
I don't think there's anything that we can do here -- at least, the dictionary can be opened in a new tab.  PR #298 changes the dictionary, that's a reasonable change for now!

PR #298, opened 2024-02-29
Fix Issue 16: clicking a word redirects to dictionary

imamcr:
**Problem**: #16

The following script runs on the ctext.org so when the dictionary is embedded it redirects to the ctext.org

```js
if (top.location != self.location) { top.location = self.location; }
```

**Solutions**
- changing ctext.org to popup
- using another dictionary

**Alternative Dictionaries**
https://en.wiktionary.org/wiki/###
https://www.archchinese.com/chinese_english_dictionary.html?find=###
https://www.mdbg.net/chinese/dictionary?page=worddict&wdrst=1&wdqb=###
*https://www.xiaoma.info/hanzi.php?hz=###

**Fix**
I changed the dictionary to archchinese.

### DIFF
--- a/lute/db/demo/languages/classical_chinese.yaml
+++ b/lute/db/demo/languages/classical_chinese.yaml
@@ -2,7 +2,7 @@ name: Classical Chinese
 dictionaries:
   - for: terms
     type: embedded
-    url: https://ctext.org/dictionary.pl?if=en&char=###
+    url: https://www.archchinese.com/chinese_english_dictionary.html?find=###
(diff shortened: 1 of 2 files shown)


jzohrab:
This is probably ok, I don't study this language so I don't know what the better alternatives are.  I agree that the existing dict is no good due to its js.  @M-Biggles has already given this the thumbs up so let's go with it.  Thank you!

jzohrab:
Launched in 3.2.4.

PR #308, opened 2024-03-08
Move vendored file to separate dirs and update templates

### LINKED ISSUE #307: Move vendored js, css into separate folder(s) in lute/static
Currently, `lute/static/css` and `lute/static/js` have some vendored code (i.e code pulled from external sources, and committed into the repo).  Vendoring is fine, but these vendored files should be stored separately.

There are a few libraries:

* jquery
* tagify
* charting
* datatables

and some of these have both js and css.  I think they should go into separate folders under vendor, eg:

```
lute/
    static/
          vendor/
                tagify/
                jquery/
... etc
```

and then the lute templates should be changed to point at the right folders.

Jayanth-Parthsarathy:
Fixes #307.
- [x] Moved the vendor files to separate folder under the vendor/ directory.
- [x] Updated the template files to use the vendor files under vendor/ directory


jzohrab (approved):
Looks good!  I'll check it out locally as well before merging.  Thank you for taking the time @Jayanth-Parthsarathy :-)

jzohrab:
Changed to develop, that's where PRs should go. :-)

jzohrab:
Launched in 3.2.5.

PR #320, opened 2024-03-11
Add SRT and VTT support

imamcr:
Added SRT and VTT support.

image
image


jzohrab:
Thank you @imamcr  !

The failed tests appear to have nothing to do with this code change.

jzohrab (changes requested):
Thank you very much!

Can you add a single sanity check acceptance test for each new file type?  The files should be added to `tests/acceptance/sample_files`, and you'll see sample tests in `tests/acceptance/book.feature`, e.g.:

```
    Scenario: I can import an epub file.
        Given I visit "/" ... etc.
```

You shouldn't need to change any code otherwise, just the tests.

Depending on your gut feeling, you may or may not want to also check for invalid files, see e.g. `Scenario: Invalid epub files are rejected.`

requirements.txt
```diff
@@ -69,6 +69,7 @@ sortedcontainers==2.4.0
 soupsieve==2.5
 splinter==0.19.0
 SQLAlchemy==2.0.21
+subtitle-parser==1.3.0
```
jzohrab, line 72: This will also need to be added to the `pyproject.toml` file, which handles dependencies separately.

jzohrab:
Thanks for the text and the .toml fix, @imamcr .

Ah, the last failing thing was a coding style problem, with python black.  The acceptance tests passed, otherwise.  I'll pull the branch and will fix that locally.

jzohrab (approved):
Looks good, thank you!

jzohrab:
Launched in 3.2.5.

PR #322, opened 2024-03-12
Book title from file

### LINKED ISSUE #272: Idea: Title from text file filename when creating new book
Simple idea: When creating a new book, you can use a text file. When you Browse and add it, perhaps the book Title can be replaced with the filename (if it's empty). For example, if the file is called "article about dogs.txt", the Title automatically changes to "article about dogs".

JS pseudo-code (works in the console), `document.getElementById('title').value=document.getElementById('textfile').value.split('\\')[2]`

Jayanth-Parthsarathy:
Fixes #272 
- [x] Automatically get title from filename if title is empty and update in form
- [x] Preserve previous form validations.
- [x] Backend code unchanged since everything is handled on the frontend

### DIFF
--- a/lute/templates/book/create_new.html
+++ b/lute/templates/book/create_new.html
@@ -95,6 +95,19 @@
 
     set_field_directions();
     $("#language_id").change(set_field_directions);
+    function getFilenameWithoutExtension(filename) {
+      return filename.split('.').slice(0, -1).join('.');
+    }
+    function autoFillTitle() {
+      var filename = $('#textfile').prop('files')[0].name;
+      var titleField = $('#title');
+      if (titleField.val() === '' && filename !== '') {
+        titleField.val(getFilenameWithoutExtension(filename));
+      }
+    }
+    $('#textfile').on('change', function() {
+      autoFillTitle();
+    });
   });
 </script>
 


lute/templates/book/create_new.html
```diff
@@ -95,6 +95,19 @@
...
+      return filename.split('.').slice(0, -1).join('.');
+    }
+    function autoFillTitle() {
+      var filename = $('#textfile').prop('files')[0].name;
+      var titleField = $('#title');
+      if (titleField.val() === '' && filename !== '') {
+        titleField.val(getFilenameWithoutExtension(filename));
+      }
+    }
+    $('#textfile').on('change', function() {
```
jzohrab, line 108: Much better than what I was thinking :-)

jzohrab (changes requested):
I don't feel this is correct -- the title should always be required, no matter what.  Let me know what you intended with this change, maybe there's something that's not clear (for you or for me).

lute/book/forms.py
```diff
@@ -18,7 +18,7 @@ class NewBookForm(FlaskForm):
 
     language_id = SelectField("Language", coerce=int)
 
-    title = StringField("Title", validators=[DataRequired(), Length(max=255)])
+    title = StringField("Title", validators=[Optional(), Length(max=255)])
```
jzohrab, line 21: (This is the section I was referring to when I said that this change felt incorrect)

lute/book/forms.py
```diff
@@ -83,7 +83,7 @@ def validate_language_id(self, field):  # pylint: disable=unused-argument
             raise ValidationError("Please select a language")
 
     def validate_text(self, field):  # pylint: disable=unused-argument
-        "Throw if missing text and textfile, or if have both."
+        "Throw if textfile is not uploaded and title is empty, or if both are provided."
```
jzohrab, line 86: This isn't correct, this is validating just text content, which is provided either with the text field or with a text file, not the title.

jzohrab: Not sure what I'm missing here :-)

lute/book/forms.py
Jayanth-Parthsarathy: Sure, this was my mistake. I first made the changes through the backend and then switched to frontend so I missed this issue. I will resolve this asap.

jzohrab (changes requested):

lute/book/forms.py
jzohrab: Cheers thanks @Jayanth-Parthsarathy :-) yeah I think this is all front-end-only.  Good idea with the js handler, much better than my bad idea!

Jayanth-Parthsarathy:
> I don't feel this is correct -- the title should always be required, no matter what. Let me know what you intended with this change, maybe there's something that's not clear (for you or for me).

I have removed the unnecessary changes from backend. The modifications on the frontend must work now. Cheers!

jzohrab:
Yes this looks good, thanks @Jayanth-Parthsarathy .  I'll pull this down to check and will and squash it to a single commit, no need for the extra thrash.  Super!

jzohrab:
Already done :-P  Thanks though!  It's still attributed to you, so that's good.  Cheers!

Closing the PR, but the code was merged manually and pushed.

jzohrab:
Launched in 3.2.5.

PR #334, opened 2024-03-14
Issue 333: Remove zero-width joiner and non-joiner characters

### LINKED ISSUE #333: Hindi word parser is confused by zero width joiner characters (U+200D)
**Description**

In the Dev Nagari writing system used by Hindi, each consonant is combined with a vowel.  "Conjunct" characters are normally used to represent sequences of consonants that do not have vowels between them, for example the "st" combinaton in "namaste".

Some Hindi web pages use a sequence of consonants with a zero-width joiner character, U+200D, in between them instead of a conjunct character.  When Lute attempts to parse words that contain these characters it splits the words and that breaks the ability to look up definitions in a dictionary.

To make things work properly, the zero--width joiner characters needs to be removed.  It is *not* enough to add this character to the list of word characters in the language.  The obvious solution would be to modify the SpaceDelimitedParser.get_parsed_tokens() method to remove zero-width joiner characters as well as zero-width space characters.

**To Reproduce**

1. Import text from the Hindi web page https://jivani.org/Biography/400/%E0%A4%9C%E0%A5%89%E0%A4%B0%E0%A5%8D%E0%A4%9C-%E0%A4%91%E0%A4%B0%E0%A4%B5%E0%A5%87%E0%A4%B2--%E0%A4%9C%E0%A5%80%E0%A4%B5%E0%A4%A8%E0%A5%80---biography-of--george-orwell-in-hindi-jivani
2. Open the boo
3. Click on the word after नामक, which looks like स्थान but contains a zero-width joiner after the first character.
4. Depending on where you click, you will either select स् or थान

**Screenshots**

None

- OS (e.g., iOS, windows): Linux
- Browser (e.g., chrome, safari): Firefox
- How you've installed Lute (Docker, python, source): pip
- Version: 3.2.5

mzraly:
Fixes #333

### DIFF
--- a/lute/parse/space_delimited_parser.py
+++ b/lute/parse/space_delimited_parser.py
@@ -27,9 +27,14 @@ class SpaceDelimitedParser(AbstractParser):
 
     def get_parsed_tokens(self, text: str, language) -> List[ParsedToken]:
         "Return parsed tokens."
+
+        # Remove extra spaces.
         clean_text = re.sub(r" +", " ", text)
-        zws = chr(0x200B)  # zero-width space
-        clean_text = clean_text.replace(zws, "")
+
+        # Remove zero-width spaces, joiners, and non-joiners.
+        for c in (chr(0x200B), chr(0x200C), chr(0x200D)):
+            clean_text = clean_text.replace(c, "")
+
         return self._parse_to_tokens(clean_text, language)
 
     def preg_match_capture(self, pattern, subject):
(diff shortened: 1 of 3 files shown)


jzohrab:
Thanks, great PR with tests and comments that helped me understand what's up.  Appreciated!

M-Biggles:
I was unable to get some languages to display correctly without these sorts of characters.

What do we do now if they are automatically removed from texts?

mzraly:
Let me try an alternate approach of adding zero-width joiners and
non-joiners to word characters so they don't break parsing but are retained
for rendering.  I'll try that tonight.

Sorry for the inconvenience.

jzohrab:
Thanks @mzraly :-)

PR #340, opened 2024-03-15
Issue 339 retain zero-width joiner and non-joiner characters

### LINKED ISSUE #339: Dropping zero-width joiners and non-joiners can break rendering for various languages
**Description**

The fix for #333 did not account for languages that require zero-width joiner and non-joiner characters to render properly.

To be decided: Fix by making these characters always considered to be parts of words by the space-separated parser, regardless of language, or just add the characters explicitly to the language definitions where needed?

**To Reproduce**

Anecdotal, but see https://github.com/w3c/i18n-discuss/issues/2#issuecomment-259986459 for how zero-width joiner can be used in Arabic and Syriac.

**Screenshots**

None

- OS (e.g., iOS, windows): Linux
- Browser (e.g., chrome, safari): Firefox
- How you've installed Lute (Docker, python, source): pip
- Version: develop

mzraly:
Fixes #339 and mostly undoes #333 by restoring original space-delimited parser behavior and explicitly adding zero-width joiner and non-joiner characters to the language definitions for German and Hindi to make unit tests pass and demonstrate how to add zero-width joiner and non-joiner support to other languages.

This is the most conservative fix that retains the existing behavior in all other languages and only modifies the word-detection behavior of German and Hindi.

### DIFF
--- a/lute/db/demo/languages/hindi.yaml
+++ b/lute/db/demo/languages/hindi.yaml
@@ -3,6 +3,9 @@ dictionaries:
   - for: terms
     type: embedded
     url: https://www.boltidictionary.com/en/search?s=###
+  - for: terms
+    type: popup
+    url: https://translate.google.com/?sl=hi&tl=en&text=###
   - for: sentences
     type: embedded
     url: https://www.bing.com/translator/?from=hi&to=en&text=###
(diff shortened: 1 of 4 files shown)


jzohrab:
Thanks, such a tricky situation.  Going forward with such changes I guess we should open PRs up for wider review to get some more input.  Cheers!

lute/db/demo/languages/hindi.yaml
```diff
@@ -13,4 +16,4 @@ show_romanization: true
 # character_substitutions:
 split_sentences: .?!|।॥
 split_sentence_exceptions: Mr.|Mrs.|Dr.|[A-Z].|Vd.|Vds.
-word_chars: a-zA-Z\u0900-\u0963\u0966-\u097F
+word_chars: a-zA-Z\u0900-\u0963\u0966-\u097F\u200C\u200D
```
jzohrab, line 19: Whoops, should have done something similar in the german.yaml file :-P  No worries, looking into it.

Issue #99, opened 2024-01-01
Change data management: always store all terms in db, even when unknown (status 0)


jzohrab:
Currently Lute doesn't actually store unknown words, they're only stored when terms are created.

Per Shetoz in Discord:

```
Hey @jz , can we add the ability to save state0 terms, so the pop-up/tool-tip shows even if the term is technically new?
Use cases for this might be:
1. currently testing a Mandarin parser with @fanyingfx, the parser adds the pronunciation automatically, but I still have to click save on each term for it to show -it's just inconvenient-
2. The more realistic one for others is, exporting pronunciation and definitions from dictionaries for the most common 1 k words or something like that, and having them on demand when you wanna read without interruption.
```

currently, Lute only stores "known words", i.e., words with status >= 1.  In Lute, "unknown words", i.e. those that show up as the light blue new terms while reading, have "status 0", and are not stored in the db.  In many ways, this makes sense: you don't know it, so it shouldn't be in the db, because it kind of doesn't exist for you yet.

Sometimes this creates problems though, or difficulties during usage:

1) when reading, if you click on an "unknown word" from the reading screen, Lute has to parse it when displaying the term form in the top right corner.  Sometimes this creates parsing inconsistencies, which are bugs, such as in issue #193.
2) fancy parsers can't auto-add some information for terms.
3) educators etc can't provide pre-populated lists of vocab, without having them create new "status = 1 (new)" terms.

The code change in the branch addresses the above this way: when a page is opened for reading, it's parsed as usual, but then Lute saves all the new "status = 0" terms it needs for that particular page.  The javascript handlers that pop up the term editing form now can use the new term's ID when opening the form, so it doesn't have to reparse the term.  This fixes point 1 above, and point 2.

Since this branch also lets people import new vocab with "import new terms as unknowns", that means that people can share pre-populated vocab lists without also overwhelming users of that data with a slew of new terms.  It's a nice benefit.

---

## Branch: `issue_99_store_new_terms_on_read`

Remaining todo items:

- [x] saving all new terms required for page render
- [x] the "WoCreated" field should change to today when the user changes the status of the term from 0 to non-0 (currently in this branch, it will be set to when the new placeholder unknown term is added. Related to #305 )
- [x] lute.js should use the word id when looking up the term in the db, not re-parsing it or doing anything strange.
- [x] perhaps multi-word-term creation on reading can change to use the word ids -- that's big change though, potentially.
[...]

PR #347, opened 2024-03-17
Issue 99: store new terms on open page for reading

jzohrab:
Addresses #99, which in turn should clear up some inconsistent parsing people were running into during reading.

Draft PR for discussion and review, but I don't know if anyone can really review it!


jzohrab:
Launched in 3.3.0.

Issue #361, opened 2024-03-22
Mobile reading: header and side margin takes up too much space


jzohrab:
From discord:

> I currently have a few annoyances with the mobile website. It has a big thick forehead/header that house the hamburger menu and page selector. This wastes a lot of vertical real estate that I need while searching dictionaries. And the 2nd biggest annoyance is there is wide borders on the left and right sides which in turn waste a lot of horizontal real estate

makes a lot of sense, should have a more mobile-friendly option.

> ![Image](https://github.com/jzohrab/lute-v3/assets/1637133/ccfaaa19-30ac-4e5c-ad96-4d9bf6508b0e)

schlick7:
![Screenshot_20240321-142137](https://github.com/jzohrab/lute-v3/assets/17865651/3cb90f95-c9d5-4fe3-81ac-c0a9e96218fa)

Here's an example screenshot

jzohrab:
Launched in 3.3.0.

PR #362, opened 2024-03-22
Issue 361: tighten up mobile css

jzohrab:
From issue #361 - the mobile css has a fair amount of whitespace which takes up real estate on smaller devices.  This PR proposes default changes to the css.  Here's a screenshot from my chrome devtools:

Before this change:

> image

After:

> image

I think some people are already using mobile so opening this up for review (by @webofpies ) and others.


jzohrab:
Launched in 3.3.0.

PR #368, opened 2024-03-23
Add autofocus to forms

imamcr:
Added autofocus to forms.


jzohrab:
Thank you @imamcr, nice fix . 👍

jzohrab:
Launched in 3.3.0.

Issue #377, opened 2024-03-26
On mobile the audio player stretches off screen and some buttons are hidden 


schlick7:
![Screenshot_20240325-202925](https://github.com/jzohrab/lute-v3/assets/17865651/d6f96f98-97fd-4309-96d6-88709fc186dc)

I did a very hacky fix on my css file which "fixes" the problem, but I doubt it works on anything else. Hoping somebody with some actual css knowledge can look at this and align things correctly. 

``
```
.audio-player-right-container {
  grid-template-columns: 14rem; /*1fr*/
  align-items: left;
  justify-items: left;
}

.rewind-container {
  grid-template-columns: min-content 1.9rem; /*1fr*/
```
Also shrunk the play button size
```
#play-btn {

  height: 48px;
  width: 48px;
}
```
![Screenshot_20240325-201931](https://github.com/jzohrab/lute-v3/assets/17865651/12f1b198-957e-4ee4-b0ef-a8573620516f)

jzohrab:
Ah right, that's better, thanks.  Maybe @webofpies can take a look if he has time, thank you very much @schlick7 for providing the detail here.

webofpies:
@schlick7 , what kind of device do you use? it looks like it's smaller than a tablet? im just curious to know if this is a bug or  the minimum player size is a little too big

webofpies:
how does this look? removed some paddings and also hid the select field altogether ![Screenshot_9](https://github.com/jzohrab/lute-v3/assets/70017511/a36fd21f-cb70-47d4-97cc-d31fac446efc)

schlick7:
I'm using a phone. Pixel 8 - 6.2" -  1080 x 2400

Your screenshot appears to have completely cut out the drop down to change how far rewind/forward goes. 

Looks pretty good otherwise. Using it on a phone will be a compromise regardless, but this looks workable. Maybe a little less padding on the outside? Those buttons are a pretty close together and would be pretty easy to miss click. Using it on a phone will be a compromise regardless, but this looks workable

jzohrab:
The changes @webofpies made in #378 look great, @schlick7 that will be in the next release.

jzohrab:
Launched in 3.3.2.

PR #378, opened 2024-03-26
Player styles for even smaller views

webofpies:
added media query for even smaller viewports (#377) image


jzohrab:
Looks great, thanks @webofpies !

jzohrab:
Launched in 3.3.2.

PR #390, opened 2024-04-03
Set focus for term form

### LINKED ISSUE #379: Set focus to translation field on open from reading screen
On open from reading screen, the term is created, most natural thing would be to edit the translation.  ref [discord](https://discord.com/channels/1074759089051160647/1155012911299833876/1222276254859001936)

Jayanth-Parthsarathy:
# Set focus properly for term form

## Fixes #379 

- [x] Add proper form focus for when creating new term, editing existing term, clicking on term in read page, etc.,

#### Clarifications thanks to @jzohrab :
- If the form is opened to create a new term and the language isn't set, focus should go there
-  If form is opened to create new and lang is set (b/c the user set the "current language" from home screen), focus goes to the term
- If lang and term are set, it should go to the translation.

### DIFF
--- a/lute/templates/term/_form.html
+++ b/lute/templates/term/_form.html
@@ -325,6 +325,19 @@
       }
     });
 
+    // Set term form focus
+    const langId = $("#language_id")
+    const text = $("#text")
+    const translation = $("#translation")
+
+    if(langId.val() == "0"){
+      langId.focus();
+    }else if(text.val() == ""){
+      text.focus();
+    }else{
+      translation.focus();
+    }
+
     // Term image events.
     var term_image = $("#term_image");
 


jzohrab (changes requested):
Hi @Jayanth-Parthsarathy , can you make a few changes?

* the code should be moved inside the `$(document).ready(function () {` function block
* for uniformity, please use jquery-style `$("#someId")` instead of `document.getElementById()`
* please use spaces for indenting :-P

Jayanth-Parthsarathy:
@jzohrab  Made the changes as you have requested. My bad for not looking at the indentations.

jzohrab:
Thanks @Jayanth-Parthsarathy, I'll try it out, looks good!

jzohrab:
Works great, thanks!

jzohrab:
Well @Jayanth-Parthsarathy -- after all your work, I'm going to disable this change, sorry about that!

Prior to this change, I could use the left and right arrow keys to move from term to term in the reading pane, and the form would update, which is nice for easily moving through terms and updating things.  With this change, the form pulls focus, so the arrow keys stop working.

Sorry again, this is one of those things where I only get a feel of the features through my own usage.  Cheers!

PR #397, opened 2024-04-08
Mention the --help flag when lute cannot start.

drm00:
Port 5000 was already taken on my machine, but besides the exception, there was no hint that the port could be set with --port.

This PR adds a hint that --help is available, where the --port-flag is mentioned. Thus, we do not document all available flags, but rather give a hint that command line flags are available.


jzohrab:
LGTM, thanks 👍

PR #400, opened 2024-04-13
Remove autofocus cruft

### LINKED ISSUE #393: Clean up term form "autofocus" cruft
When PR #390 is merged, there is some cruft to clean up:

```
# I think the js from the above PR overrides this
lute/templates/term/_form.html:      <div>{{ form.text(class="form-control", value=(form.original_text.data or ''), autofocus=(not form.original_text.data)) }}</div>

# I  think the "{autofocus}" extra  args can be removed, it's not used
lute/static/js/lute.js:  show_term_edit_form(target, { autofocus: false });

# I'm pretty sure none of this works
lute/templates/term/_form.html:  // TODO zzfuture fix: check term autofocus
lute/templates/term/_form.html:  // lute.js should send an "autofocus" flag, I believe.
lute/templates/term/_form.html:    if ($('#autofocus').val() != 'false') {
```

Jayanth-Parthsarathy:
### Fixes #393 

- [x] Removed all the specified instances of auto-focus specified in the issue #393
- [x]  Checked if auto-focus works with these code removed.

### DIFF
--- a/lute/static/js/lute.js
+++ b/lute/static/js/lute.js
@@ -369,7 +369,7 @@ let move_cursor = function(shiftby) {
   save_curr_data_order(target);
   apply_status_class(target);
   $(window).scrollTo(target, { axis: 'y', offset: -150 });
-  show_term_edit_form(target, { autofocus: false });
+  show_term_edit_form(target);
 }
 
 
--- a/lute/templates/term/_form.html
+++ b/lute/templates/term/_form.html
@@ -33,7 +33,7 @@
   
       {{ form.original_text }}
 
-      <div>{{ form.text(class="form-control", value=(form.original_text.data or ''), autofocus=(not form.original_text.data)) }}</div>
+      <div>{{ form.text(class="form-control", value=(form.original_text.data or '')) }}</div>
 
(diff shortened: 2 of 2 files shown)


jzohrab:
Thank you @Jayanth-Parthsarathy , I'll check it out later!  👍

lute/static/js/lute.js
```diff
@@ -369,7 +369,6 @@ let move_cursor = function(shiftby) {
   save_curr_data_order(target);
   apply_status_class(target);
   $(window).scrollTo(target, { axis: 'y', offset: -150 });
-  show_term_edit_form(target, { autofocus: false });
```
jzohrab, line 372: Hi @Jayanth-Parthsarathy - 99% sure this line is still needed, just the `{ autofocus: false }` part should be removed ... right?  The form should be shown if the user clicks a term.

jzohrab (changes requested):
Single change requested.

jzohrab:
Thanks @Jayanth-Parthsarathy , I'll give this a shot soon.  Cheers!

jzohrab (approved):
Thank you!

jzohrab:
Launched in 3.3.2.

PR #366, opened 2024-03-23
Issue 349: Provide default values for sentence terminators and word characters

mzraly:
This change for #349 lets users of the space-delimited parser leave the "Split sentences at" and "Word characters" language attributes empty.  This lets users define new space-delimited languages without worrying about those fields.

What changed:
* The language form no longer validates that these fields have data.
* The labels for these fields now describe the default values.
* The space-delimited parser now handles an empty "Split sentences at" field by treating all Unicode sentence terminators, plus the colon character, as sentence terminators.
* The space-delimited parser now handles an empty "Word characters" field by treating all Unicode letters,marks, and format codes as word characters.
* Added a new language definition called "Generic"
* Added a demo story for language Generic with sample text from a variety of languages and writing systems.
* Added unit tests for parsing using the Generic language.

What didn't change?
* Did not add a dependency on the regex package, as I originally planned.
* Did not change the definition of any existing language.
* Did not add default values for the "Character substitutions" field.
* Did not add default values for the "Split sentence exceptions" field.
* Did not change the behavior of the space-delimited parser with respect to embedded punctuation characters like hyphens and apostrophes.

PR #407, opened 2024-04-28
Issue 349 default wordchars rebased

jzohrab:
Replacement for PR #366 

Rebased the original PR off of develop, which had changed after that PR was opened, and then refactored method that built the unicode character range for the word chars regex.


mzraly (approved):
Looks good to me.  I agree that your refactoring of get_default_word_characters makes it clearer.

PR #413, opened 2024-05-10
Enable optional plugins for parsers through a project entrypoint; Add Optional Mandarin Parser Plugin

cghyzel:
### Summary of Changes

- Added  optional plugin initialization logic to `lute/parse/register.py`
- Added plugin initialization and logging to `lute/main.py`
- Refactored lute/parser/register.py to use `__LUTE_PARSERS__` instead of `parsers` for the module wide variable.
- Added a Mandarin Chinese Parser Plugin in `plugins/lute-mandarin` (This will require a separate installation by the user).
- Added relevant tests to plugins/lute-mandarin/tests
- Added jieba (https://pypi.org/project/jieba/) to requirements for `plugins/lute-mandarin` for parsing because Mandarin Chinese has no clear word delimiters and so it is difficult to parse. It is under the MIT License.
- Added pypinyin (https://pypi.org/project/pypinyin/) to requirements for `plugins/lute-mandarin`. It is used for readings of Chinese characters.It is under the MIT License.
- Updated the version to `3.4.0` and required `lute>=3.40` for the lute-mandarin plugin.

When I had submitted [my original pr for a mandarin parsers](https://github.com/LuteOrg/lute-v3/pull/408), it nearly doubled the package size due to the dependencies on Jieba and pypinyin. Now lute mandarin can be installed separately and encorporated automatically thanks to the addition of the `init_parser_plugins()` method in `lute/parse/registry.py`

When installed it will automatically be picked up by lute. 

From root just run:
```
pip install .
pushd plugins/lute-mandarin
 pip install . 
 popd
 python -m lute.main --port 9876 --config lute/config/config.yml
 ```

When the package is picked up it will show in the initialization log like so:
```
Initializing parsers from plugins...
  Enabled Parsers:
  * (Space Delimited)
  * (Turkish)
  * (Classical Chinese)
  * (Lute Mandarin Chinese)
```

When this work is merged a new package for lute-mandarin can be added to PyPi and installed by users who want to use lute to learn Mandarin. This way there is no additional package bloat. Parsers for other languages that don't fit into the model of the space delimited parser like Korean and Czech will also be able to leverage the new entry point and can depend on additional libraries without that same concern for package bloat to the main package.

### CODE REVIEW

**Comment by jzohrab** [comment]
Thanks very much for the first pass at this problem!

I've read through the code and the docs you added, it makes sense as a git-only installation, but I'd like to keep this as pip plugins which augment the base install, instead of full installations on their own, if that makes sense.  Also, most users don't have git :-)

I much prefer the idea you mentioned [in Discord](https://discord.com/channels/1074759089051160647/1235323613465415730/1237489527086383215): "I need to learn Korean => pip install lute-korean-parser =>  run lute ".  I had what feels like roughly the same idea, and wrote this before reviewing that comment :-) :

```
mkdir -p ~/my_lute
cd ~/my_lute
python3 -m venv myenv
source myenv/bin/activate
pip install lute3   # the main lute
pip install lute3-mandarin-chinese    # add the new plugin!
python -m lute.main
```

This way the plugin would be a minimal increment on top of the existing code, instead of being included in the source code.

As you mentioned, the link https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/ has some notes about how we can do this.  The simplest option appears to be to use a naming convention, on first read through.  There are a few things I'm not clear how to manage with the plugin approach:

- **class heirarchy:** Currently Lute parsers use an AbstractParser as a base class, to ensure that parsers conform to it.  Sure python has duck typing and so base classes etc aren't _strictly_ required, but it seems like a good idea to have that.  Since each parser would be in its own repo, how would they get access to the AbstractParser base class to inherit from?  Would they have `lute3` in their own requirements.txt (should be fine, but depending on lute versioning could create weird dependency problems)?  Or would there need to be some kind of `lute-interfaces` repo/package that both Lute and any parsers include?
- **language definitions:** if a given language X needs its own special parser, then presumably the definition for that language will refer to that parser!  Currently, language definitions are stored in the `lute/db/language-defs` git submodule ... but then that could create a situation where a language-def and sample story exists for language X in the lute language listing, but the parser doesn't exist.  Should the language-def and sample story be included as part of the plugin itself?  Part of me thinks "yes", because if I make a plugin for Korean, I might want to submit the stories etc for it too.  Of course, this means that the `definition.yml` for the language would need to meet the requirements for Lute as well :-)
- **package compat checks:** (perhaps a minor problem!) if packages include many deps, there may be conflicts.  Of course packages should minimize the number of deps they install -- eg for a mandarin parser it might only add jieba and pypinyin
- **installation:** for python/pip users, I think the above `pip install` is totally fine. :-)  For Docker users, though, I'm not sure, and I think we can skip that for now as Docker users comprise a smaller segment of the user pop ... docker is a pre-baked image, maybe we'd just do docs for them to actually ssh into their containers, or docs for them to build their own containers with whatever plugins they need.

I have some cycles over the next few days, and might mess around with some fake packages in test pypi to see what I come up with.  But happy to chat about it!

Small notes:

- prefer topic branches off of `develop`, rather than using your `master` as an ongoing branch.  Reason: this PR contains commits from the original Mandarin branch.  (You may know this already, as this is a PR for discussion.)
- Release number changes happen during actual release process.  Other things might be included in the release.  Thanks though!

**Comment by cghyzel** [comment]
Hey @jzohrab , thanks for reviewing this in detail!

### The User Story

The original idea is actually already implemented here, if you `python3 -m build` in the `plugins/lute-mandarin` directory it creates a plugin that can be uploaded to PyPi. This works because  `plugin/lute-mandarin` directory has  `pyproject.toml` with separate dependencies. Users can then run `pip install lute-mandarin` and then the Mandarin parser will be mounted when they run lute automatically. 

### Package Dependency
Plugins can add `lute` as a dependency to free them from managing all but their essential and unique dependencies, like `jieba` and `pypinyin` for lute-chinese`. There reason I upgraded the lute version in my PR is because the packaging instructions don't work otherwise, that's because I added a version specific dependency on lute to `plugins/lute-mandarin/pyproject.toml`:
```
dependencies = [
  "lute3>=3.4.0",
  "jieba>=0.42.1",
  "pypinyin>=0.51.0"
]
```
I added the dependency on a later version of lute because the current lute version doesn't actually support the plugin, so it felt illogical to keep the lower version. You can definitely change this to your liking. 

### On the Class Hierachy
Because lute is included in the dependencies the plugin can do things like this, just like you described, to maintain the class hierarchy. 
```
from lute.parse.base import ParsedToken, AbstractParser

class MandarinParser(AbstractParser):
```
and then that AbstractParsers is actually validated as during the plugin initialization
```
        if _is_valid(custom_parser_ep.load()):
            __LUTE_PARSERS__[custom_parser_ep.name] = custom_parser_ep.load()
        else:
            raise ValueError(f"{custom_parser_ep.name} is not a a subclass of AbstractParser")
```
... and the actual implementation of is_valid.
```
def _is_valid(custom_parser):
    return issubclass(custom_parser, AbstractParser)
```
### Language Definitions
I included a language definition in the plugin base directory and it is used to build the Language object in `plugin/lute-mandarin/tests/conftext.py`.
```
def _get_test_language():
    """
    Retrieve the language definition file for testing ths plugin from definition.yaml
    """
    thisdir = os.path.dirname(os.path.realpath(__file__))
    definition_file = os.path.join(thisdir, "..", "definition.yaml")
    with open(definition_file, "r", encoding="utf-8")  as df:
        d = yaml.safe_load(df)
    lang = Language.from_dict(d)
    return lang
  ```
As for how to include language definitions for each plugin, we would either need an injection method for language definitions from the plugin module, or you could continue to use the `lute/db/language-defs` but simply hide language definitions that don't have a supported parser available. Does handling the language definition block the PR, or could the problem be addressed separately?

jzohrab:
Thanks for all the notes and for thinking through everything.  My code review was obviously lacking!  I missed all of the important details in the code, such as versioning etc.  :-(  Ah well, cheers.

- "... reason I upgraded the lute version in my PR": makes perfect sense.  We could get pedantic and release the new version with the capability first, and then the new plugin afterwards.  That feels like make-work -- however it may uncover an implicit dependency or assumption that we've missed.  I'll change the version to beta and do a test release in test pypi for this PR and also for the plugin - I did that in the past (ref https://test.pypi.org/search/?q=lute3).
  
- "simply hide language definitions that don't have a supported parser available": the code already does that actually.  `lute.parse.registry.is_supported(parser_name)` ([see code](https://github.com/LuteOrg/lute-v3/blob/master/lute/parse/registry.py#L40)) checks the array (now `__LUTE_PARSERS__`) to see if the parser exists, and then if it's really supported.  Some parsers like Japanese need extra stuff installed (mecab) to actually be viable.  Side note: I need to fix this somehow, b/c it's not _really_ working well, but that's stuff for another day.

## language-defs

"Does handling the language definition block the PR, or could the problem be addressed separately?" -- good question,  I don't think it blocks it.

The current organization of data in `lute/db/language-defs/` feels wrong, i.e. the definition.yml files in one arbitrary directory could refer to plugins found elsewhere.  The PR adds a definition.yml for the language for its tests, which is the right thing to do for cohesion.

The existing Lute code in `lute.language.service.predefined_languages()` reads the language defs just from the `lute/db/language-defs/`.  If I install the plug-in with pip, then the language should appear in the listing.  Right now, I don't think it will, unless we have a copy of the language and its stories in language-defs.  Having the same yaml file in two locations is kludgy but not terrible, i.e. one copy here, and another copy in the language-defs.

My feeling is that the plugin should provide its own language definition and stories, as you did here with the definition.yml.  The `lute.language.service` could get the language defs and stories from wherever `language-defs` are kept, _plus_ by querying the plugins.  That can be done later.

# dependencies

This looks like you did "pip freeze > requirements.txt" to get everything, but that also pulled in all of the stuff that lute itself installed, meaning that we may end up with conflict clashes between the projects if both are installed into the same venv.  I'll see about reducing the dependencies in my test branch I use to publish to testpypi.

lute/parse/registry.py
```diff
@@ -4,57 +4,69 @@
...
 from lute.parse.base import AbstractParser
 from lute.parse.space_delimited_parser import SpaceDelimitedParser, TurkishParser
 from lute.parse.mecab_parser import JapaneseParser
 from lute.parse.character_parser import ClassicalChineseParser
 
 
-# List of ALL parsers available, not necessarily all supported.
-# This design feels fishy, but it suffices for now.
-parsers = {
+__LUTE_PARSERS__ = {
```
jzohrab, line 15: Any reason for the change ?  I don't disagree, I just don't follow the motivation.  More expressive/indicative of its intent?  (I'm not a big fan of how the parsers array is managed.)

jzohrab:
I did a test release of this as **3.4.0dev1** to testpypi, it's at https://test.pypi.org/project/lute3/3.4.0.dev1/.  I'll try an install of that to see what happens.

I installed the plugin requirements (jieba, pypinyin, lute3==3.4.0dev1) into a new .venv just to see what would happen:

```
deactivate
rm -rf .venv
python3 -m venv .venv
source .venv/bin/activate
cd plugins/lute-mandarin/
# install reqs -- get lute3 dev release from test
pip install jieba
pip install pypinyin
pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ lute3==3.4.0.dev1
pip freeze > requirements.txt 
```

and the resulting plugins/lute-mandarin/requirements.txt is a fair bit smaller (38 lines, vs 91 lines).  Below is a slightly processed diff, removing version number changes and case changes -- there are some errors in the processing but it's mostly correct:

<details>
<summary>diff requirements file</summary>

```
diff --git a/plugins/lute-mandarin/requirements.txt b/plugins/lute-mandarin/requirements.txt
index c6123e98..516cb683 100644
--- a/plugins/lute-mandarin/requirements.txt
+++ b/plugins/lute-mandarin/requirements.txt
@@ -1,91 +1,38 @@
-astroid==2.15.6
-attrs==23.1.0
-black==23.10.1
 cffi==1.16.0
-cfgv==3.4.0
+chardet==5.2.0
 click==8.1.7
-colorama==0.4.6
-coverage==7.3.1
-dill==0.3.7
-distlib==0.3.7
-docutils==0.20.1
-exceptiongroup==1.1.3
-filelock==3.13.1
-flit==3.9.0
-flit_core==3.9.0
-h11==0.14.0
-identify==2.5.31
-iniconfig==2.0.0
-invoke==2.2.0
-isort==5.12.0
 jaconv==0.3.4
 jieba==0.42.1
-lazy-object-proxy==1.9.0
-Mako==1.2.4
-MarkupSafe==2.1.3
-mccabe==0.7.0
-mypy-extensions==1.0.0
+lute3==3.4.0.dev1
+MarkupSafe==2.1.5
 natto-py==1.0.1
-nodeenv==1.8.0
 openepub==0.0.8
-outcome==1.3.0.post0
-packaging==23.1
-parse==1.19.1
-parse-type==0.6.2
-pathspec==0.11.2
-pipdeptree==2.13.0
-playwright==1.39.0
-pluggy==1.3.0
-pre-commit==3.5.0
-pyee==11.0.1
-pylint==2.17.5
-PySocks==1.7.1
-pytest==7.4.2
-pytest-base-url==2.0.0
-pytest-bdd==7.0.0
-pytest-playwright==0.4.3
-pytest-splinter==3.3.2
-python-slugify==8.0.1
-selenium==4.14.0
-six==1.16.0
-sniffio==1.3.0
-sortedcontainers==2.4.0
-splinter==0.19.0
-text-unidecode==1.3
-tomli==2.0.1
-tomli_w==1.0.0
-tomlkit==0.12.1
-trio==0.22.2
-trio-websocket==0.11.1
-virtualenv==20.24.6
-wrapt==1.15.0
-wsproto==1.2.0
```

</details>

@cghyzel - Part of me thinks that we should do one release of lute without the plugin, and then use that released version of lute as the source of the plugin in the requirements.  But this might be overkill.  Thoughts?

jzohrab:
All right!  New lute with plug-in capability released to testpypi, and the plugin to testpypi as well as its own package, both installed separately, and everything seems to work!  :tada: Awesome stuff, @cghyzel .

## How I tested it

### Releasing to test pypi

Releasing lute itself from main directory:

```
flit publish --repository testpypi
```

Releasing the parser:

```
pushd plugins/lute-mandarin/
pip install -r requirements.txt
# message received: Successfully installed jieba-0.42.1 pypinyin-0.51.0
flit publish --repository testpypi
```

It's up on https://test.pypi.org/project/lute-mandarin/


### First installing Lute, without the parser:

```
# created a new folder, cd'd to it
python3 -m venv .venv
source .venv/bin/activate
pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ lute3==3.4.0.dev1
# created a config ...
# started it
python -m lute.main

  Starting Lute:
  Using config.yml found in root
  
  Initializing parsers from plugins...
  Enabled Parsers:
  * (Space Delimited)
  * (Turkish)
  * (Classical Chinese)
```

All works fine.

(manually adding a file and lang def to the package)

```
$ ls .venv/lib/python3.9/site-packages/lute/db/language_defs/mandarin/
definition.yaml	story_1.txt

$ cat .venv/lib/python3.9/site-packages/lute/db/language_defs/mandarin/definition.yaml 
name: Mandarin Chinese
dictionaries:
  - for: terms
    type: embedded
    url: https://chinese.yabla.com/chinese-english-pinyin-dictionary.php?define=###
  - for: sentences
    type: popup
    url: https://www.deepl.com/translator#ch/en/###
show_romanization: true
# right_to_left:

parser_type: lute_mandarin
# character_substitutions:
split_sentences: .!?。！？
# split_sentence_exceptions:
word_chars: 一-龥
```

### Installing the mandarin plug-in

Doing an install of the 0.0.1dev1 plugin in the new folder, with the venv activated:

```
# stop Lute, but don't deactivate the venv, then:
pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ lute-mandarin==0.0.1.dev1
# message: Successfully installed jieba-0.42.1 lute-mandarin-0.0.1.dev1 pypinyin-0.51.0.1

  Initializing parsers from plugins...
  Enabled Parsers:
  * (Space Delimited)
  * (Turkish)
  * (Classical Chinese)
  * (Lute Mandarin Chinese)
```

And I can now add Mandarin as a predefined language, and the story is there.  Super!

## Notes/questions/etc

* Well, this is awesome.
* Add correct language def and story to the language-defs repo.
* I don't know if we're going to run into any package version clashes between the two packages.  Lute's packages aren't explicitly pinned in the flit release -- I'm actually not even sure how to do that.  Perhaps we just wait and see.
* I don't think it's a useful exercise to split the mandarin parser into its own independent repo.
* We should run the mandarin parser unit tests as part of CI -- but I'm currently not sure how to do that :-)  I tried various permutations of commands but got variants of `ModuleNotFoundError: No module named 'lute'` -- any guidance would be great.
* Need to add notes to the "releases" section about doing pre-checks and releases for the plugins.  Pretty much already done above
* Add some docs to the wiki about how to make parsers for other languages
* Add notes to the manual about how to find and install new parser plugins
* Sort out how to do this for Docker -- maybe with Docker we just install _everything_, b/c it's already pretty heavyweight.  I currently have a "lute" and "lute-lean" pre-built images (the latter without mecab and dictionaries), but might scrap the "lean" and just go with "lute" with everything.

lute/parse/registry.py
cghyzel: @jzohrab It makes the code easier to read because the module wide variable is obvious, and it also make classes importing the variable directly look odd, because it they should use a separate access method to prevent any unintended runtime changes to the parsers list.

cghyzel:
> @cghyzel - Part of me thinks that we should do one release of lute without the plugin, and then use that released version of lute as the source of the plugin in the requirements. But this might be overkill. Thoughts?

I don't know enough about the release process to say for sure, but first releasing lute main then the mandarin plugin sounds logical to me. You can probably just do the lute release as normal and then do a first release of the plugin. 

> We should run the mandarin parser unit tests as part of CI -- but I'm currently not sure how to do that :-) I tried various permutations of commands but got variants of ModuleNotFoundError: No module named 'lute' -- any guidance would be great.

You can add a `needs` step before the lute-mandarin integration tests are run, to be 

```
name: test plugins
needs: Verify flit install
...
run: |
for plugin in $(ls plugins); do
cd plugins/$plugin 
# install
pip install .
# Test
pytest tests
# uninstall
pip uninstall $plugin -y
done
```
This assumes the plugin name matches the directory name of course.

lute/parse/registry.py
jzohrab: Makes sense, thanks.

jzohrab:
Noticed a few things while working through this PR.  I'll fix them as I go, as some of the existing Lute code isn't that great, and this PR helps me clean it up.  I'm working on a local branch but will push it to this repo soon, even if it's incomplete.

## Un-parseable languages are shown in the "add language" listing

The "mandarin chinese" language shows up in the list of languages when clicking "add predefined language", even if the parser plugin isn't loaded.  This is because Lute's `language.service.get_defs()` returns all the languages for which definition files are present, even if the language doesn't have a valid parser available.  `mandarin chinese` should be added to Lute's predefined defs even if the plugin isn't there (at least, according to the current way of managing this data), so Lute's code needs to change to hide langs for which no parser is available.

This bug shows yet again that the `parse.registry` code is hard to grok.

## App factory needs to load plugins

Currently the plugins are loaded in `lute/main.py`, which is the prod setup, but not the `devstart.py`, which is the development setup.  I think that the plugins should be loaded as part of the application factory.

Side note: The code in `main.py` is quite messy, so I should clean that up, and perhaps clean up part of `devstart` too -- it's easy to see how this would have been missed during dev.  Maybe `devstart` can be removed ... that's another issue.

jzohrab:
I think the above points (unparseable langs shown in listing, and plugins not loaded in dev) are fixed in the branch `pr_413_continued_plugins`, pushed to this repo.  @cghyzel , I've pushed a handful of commits on top of the branch for this PR: https://github.com/LuteOrg/lute-v3/commits/pr_413_continued_plugins/ 

Rough summary:

- fixed some broken tests, black formatting
- moved plugin loading to app factory
- some cleanup of `devstart.py` and `lute/main.py`
- added the final mandarin language def and story to the language-defs submodule
- renamed plugin pypi name to `lute3-...`

Next thing to do: bump the dev release numbers again for both of these things and will push them to test pypi, reinstall in a fresh environment, and see that everything works as expected:

* bump dev release nums
* release of branch to test pypi, parser to test pypi
* install lute, no plugin, in fresh environment
* chinese mandarin should not be visible in listing
* create a new lang, mand. lang should not be vis in choice
* stop lute, install mand. plugin, restart
* see parser loaded at startup
* lang is shown in listing of pre-configured langs (don't install)
* parser can be selected as parser for new language
* do actual install of lang and story
* try it out

jzohrab:
Changing this to a Draft PR so I don't accidentally merge it.  I'll eventually merge the branch `pr_413_continued_plugins` instead, but keeping this open as it has some useful notes.

jzohrab:
Woot, merged the other branch, which also contained these commits, into the develop branch.  Will complete the various developer and user docs (or at least write them up and then iterate as usual).

jzohrab:
@cghyzel - in case you wanted to know about the tweaks done to get this into the `develop` branch etc, you could fetch this repo, and then look at the extra commits: `git log 7fe027b5..6fa71272 --oneline`, after your final `use a tuple for supported_parsers` commit.  Just little things.

* Added wiki docs: https://github.com/LuteOrg/lute-v3/wiki/Developing-language-parser-plugins , https://github.com/LuteOrg/lute-v3/wiki/Releasing-language-parser-plugins
* Added user docs: https://luteorg.github.io/lute-manual/install/plugins.html

It should be good enough as a start.  :tada: Super work again, thanks to you and to whoisnick :wave:

Issue #419, opened 2024-05-14
Add "interaction type" to force mobile/desktop js handlers


jzohrab:
The current `_isUserUsingMobile()` func in lute.js is not working for some.  Some people have interesting setups with touchscreens and monitors, and it's hard for the js to get it exactly right.  Probably the best solution is to let users explicitly toggle between interaction types in the reading slide-out menu.

This could be stored in local storage, just like the font sizes etc, and then consulted when determining how it should behave.

Spec:

* if the local storage is not explicitly set, use the value returned by `_isUserUsingMobile()`
* if set, use it
* if changed, store it, and use it

PR #421, opened 2024-05-15
Add mobile mode button in the slide-out menu

lnfu:
#419 

- Added a mobile mode toggle button in the slide-out menu.
- Modified the `_isUserUsingMobile()` function in `lute.js` to prioritize checking the local storage for a previously set mobile mode value (boolean).

### DIFF
--- a/lute/static/css/styles.css
+++ b/lute/static/css/styles.css
@@ -3,7 +3,7 @@
  * \brief Main stylesheet for the default theme.
  */
 
-* {
+ * {
     margin: 0;
     padding: 0;
 }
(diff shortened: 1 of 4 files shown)


jzohrab:
Thank you, @lnfu !  I hope to try this out today.  Really appreciated, this has been bothering some users. Cheers!

jzohrab:
Hi @lnfu , thanks again for this PR.  After some thought, I've gone in a slightly different direction, which is now in the `develop` branch:

> image

There were a few reasons for this change:

* the "focus" toggle at the top is kind of special, and I felt that adding another toggle at the top might be confusing, and add clutter.
* _most_ people shouldn't need to use the new toggle, so I moved it to the bottom of the list of items

As a dev myself, I've always found it discouraging when my PRs are declined, but I hope that the above makes sense to you.  I appreciate that you invested time into making this fix, so as a fellow dev, cheers, and I hope that you'll continue poking around at Lute's source code.  Regards, jz

Issue #424, opened 2024-05-18
Failure to start Lute 3.4.0 on Python 3.12 


cghyzel:
**Description**
(from discord)
When running on Python 3.12 the following error occurs.

```
python -m lute.main

  Starting Lute.

  Using default config
    data path: /home/ali/.local/share/Lute3
    database:  /home/ali/.local/share/Lute3/lute.db

  Initializing app.
  Initializing parsers from plugins ...


--------------------------------------------------
Error during startup:
'EntryPoints' object has no attribute 'get'
Please try again, or report an issue on GitHub.
Additionally, help is available with --help.
-------------------------------------------------
```

**To Reproduce**

Start lute on python 3.12 on version 0.3.4.0

**Extra software info, if not already included in the Description:**

Python version 3.12.1

cghyzel:
One implicit aspect of this bug is that there actually needs to be an increase in the scope of testing to encompass 3.12. Right now testing is done for 3.8 to 3.11.

thangnguyenxtz:
I can run Lute 3.4.0 with Python version 3.11, but I can't run it with Python version 3.12 too

cghyzel:
This fixed worked for me in `3.12`. Going to open a PR.
```
def init_parser_plugins():
    """
    Initialize parsers from plugins
    """
    if sys.version_info.major == 3 and sys.version_info.minor in (8, 9, 10, 11):
        custom_parser_eps = entry_points().get('lute.plugin.parse')
    elif sys.version_info.major == 3 and sys.version_info.minor == 12:
        custom_parser_eps = entry_points().select(group='lute.plugin.parse')
    else:
        custom_parser_eps = []
    for custom_parser_ep in custom_parser_eps:
        if _is_valid(custom_parser_ep.load()):
            __LUTE_PARSERS__[custom_parser_ep.name] = custom_parser_ep.load()
        else:
            raise ValueError(f"{custom_parser_ep.name} is not a a subclass of AbstractParser")
```

cghyzel:
Should be solved by 
- https://github.com/LuteOrg/lute-v3/pull/425

jzohrab:
Thanks very much @cghyzel -- merged into `develop`, and github ci was extended to include python 3.12.

PR #425, opened 2024-05-19
fix init_parser_plugins for python 3.12

cghyzel:
Hotfix for users on python 3.12, which broke because the entry points no longer return as a dictionary in that version, so ` .select(group='lute.plugin.parse')` is used instead. 

Additional logic was added to handle version differences, both in terms of fetching the entry points from the method and processing them (a catch for None was added)

**Solves:**
- https://github.com/LuteOrg/lute-v3/issues/424

### DIFF
--- a/lute/parse/registry.py
+++ b/lute/parse/registry.py
@@ -5,6 +5,7 @@ List of available parsers.
 """
 
 from importlib.metadata import entry_points
+from sys import version_info
 
 from lute.parse.base import AbstractParser
 from lute.parse.space_delimited_parser import SpaceDelimitedParser, TurkishParser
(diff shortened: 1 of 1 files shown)


lute/parse/registry.py
```diff
@@ -24,13 +25,21 @@ def init_parser_plugins():
     """
     Initialize parsers from plugins
     """
-    custom_parser_eps = entry_points().get("lute.plugin.parse", [])
+
+    if version_info.major == 3 and version_info.minor in (8, 9, 10, 11):
+        custom_parser_eps = entry_points().get('lute.plugin.parse')
+    elif version_info.major == 3 and version_info.minor == 12:
```
jzohrab, line 31: Perhaps we should change this to `>= 12`.  Granted, we can't know what the API will be in version 3.13+, but maybe we can play fast and loose with this one, and just bump up the CI checks on GitHub as versions come out.

jzohrab, line 31: Perhaps we should change this to `>= 12`.  Granted, we can't know what the API will be in version 3.13+, but maybe we can play fast and loose with this one, and just bump up the CI checks on GitHub as versions come out.

lute/parse/registry.py
```diff
@@ -24,13 +25,21 @@ def init_parser_plugins():
...
+        custom_parser_eps = None
+    if custom_parser_eps is None:
+        return
     for custom_parser_ep in custom_parser_eps:
         if _is_valid(custom_parser_ep.load()):
             __LUTE_PARSERS__[custom_parser_ep.name] = custom_parser_ep.load()
         else:
             raise ValueError(
-                f"{custom_parser_ep.name} is not a subclass of AbstractParser"
+                f"{custom_parser_ep.name} is not a a subclass of AbstractParser"
```
jzohrab, line 42: I removed the "a a" before, and here it's added again, any reason? :-)

jzohrab (changes requested):
Thank you very much!  You've probably gone to bed so I'll take the PR and make the changes to it, and then merge it.  Appreciated!

jzohrab:
Code included in updated #426 , cheers!

Issue #86, opened 2023-12-29
Add "annotated bookmarks"


jzohrab:
Lute has multi-page books.  It would be nice to add "bookmarks" to pages, for a few reasons:

* it makes sense, you can do this in a Kobo, why not in lute?
* a page might have a nice construction you want to review later
* mark chapters, etc
* Currently you can go to the prev or next page, or jump by 10, but it would be nice to go to an arbitrary page, per request #29.  But you'd want to return to where you were ...

Adding annotated bookmarks would let you add a bookmark, plus a brief optional note about what that bookmark is for.  Then maybe you could see a listing of bookmarks, and click it to go to the bookmark.

Implementation notes:

* db change - this could either be a new "TxBookmark" nullable text field in the `texts` table, or a separate table `textbookmarks` with fields `TbID, TbTxID, TxBookmark` ... the latter makes more sense db design wise, because most text records will not be bookmarked, but joins and deletes etc are a hassle so just adding `TxBookmark` would be fine by me.
* add "add bookmark" menu item in the slideout, and potentially "update bookmark" to update annotation, and "remove bookmark"
* have a "show bookmarks" page, which shows the annotations and links and page number
* a page with a bookmark could potentially show a "bookmark" image in the top right, to indicate that it's bookmarked

cblanken:
@jzohrab I've been looking into this issue since I'd really like to have a bookmark feature. I agree that a separate `textbookmarks` table is probably the way to go. Especially if it ever needs to be expanded later on. Do you have anything in the works for this, or should I be good to try my hand at it?

jzohrab:
Hi @cblanken - nothing in the works for this.

I was leaning towards _not_ having a separate table, just to reduce join hassles (b/c sometimes doing these mappings with sqlalchemy is a pain, and having separate tables means that we need to ensure deletion of the other records etc etc).  If we add a nullable bookmark field to the `texts` table we're fine, and since it will be much less code around the object management, I think that's a much better idea. :-)  LMK if you have any hard disagreements on that one, maybe there is something I'm missing.

This feature adds a bunch of operations for bookmark management to the UI as noted above.  I wonder if there is a clever way that we can break this feature down so that it can be incrementally delivered ... Maybe we do something like this, to split up the work too if needed:

## Part one: all backend work

* add field to `texts` table, necessary tests (e.g. domain to db object mapping, if any!)
* add routes for adding/editing/removing book bookmarks

[...]

PR #439, opened 2024-06-07
Add text bookmarks

cblanken:
This PR addresses issue #86 to add text bookmarks.

@jzohrab I'm pretty sure I've hit all the main requirements for the backend and smoke tests. Let me know if there's anything else I need to add or fixup before it would be ready to merge.

I think my next steps will probably be to add some demo data and maybe a mention of bookmarks in the Tutorial / Tutorial follow-up chapter and some basic acceptance tests.

I just wanted to get this PR posted before I lost steam :)

### DIFF
--- a/lute/bookmarks/routes.py
+++ b/lute/bookmarks/routes.py
@@ -52,7 +52,7 @@ def add_bookmark():
         .filter(Text.order == page_num)
         .first()
     )
-    bookmark = TextBookmark(title=title, tx_id=tx.id)
+    bookmark = TextBookmark(title=title, text=tx)
 
     db.session.add(bookmark)
     db.session.commit()
@@ -65,25 +65,22 @@ def delete_bookmark():
     data = request.json
     title = data.get("title")
     try:
-        text_id = int(data.get("text_id"))
+        bookmark_id = int(data.get("bookmark_id"))
     except ValueError:
         return jsonify(
             success=False,
-            reason=f"Invalid Text ID ({data.get('text_id')}) provided.",
+            reason=f"Invalid Text ID ({data.get('bookmark_id')}) provided.",
             status=200,
         )
 
-    if not title or not text_id:
+    if not title or not bookmark_id:
         return jsonify(
             success=False,
-            reason="Missing value for required parameter 'title' or 'text_id'.",
+            reason="Missing value for required parameter 'title' or 'bookmark_id'.",
             status=200,
         )
 
-    db.session.query(TextBookmark).filter(
-        TextBookmark.title == title,
-        TextBookmark.text.has(id=text_id),
-    ).delete()
+    db.session.query(TextBookmark).filter(TextBookmark.id == bookmark_id).delete()
     db.session.commit()
     return jsonify(success=True, status=200)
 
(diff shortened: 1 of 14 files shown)


jzohrab:
Thank you very much @cblanken !  I hope to start review today.

lute/bookmarks/routes.py
```diff
@@ -0,0 +1,114 @@
...
+            status=200,
+        )
+
+    tx = (
+        db.session.query(Text)
+        .filter(Text.bk_id == book_id)
+        .filter(Text.order == page_num)
+        .first()
+    )
+    bookmark = TextBookmark(title=title, tx_id=tx.id)
```
jzohrab, line 55: I wonder if this should use the `tx` Text object, instead of the `tx.id`.  I don't know what the best practice is -- at the code level objects are more descriptive than IDs, but it's such a minor point.  No preference, just a question.

lute/bookmarks/routes.py
```diff
@@ -0,0 +1,114 @@
...
+        return jsonify(
+            success=False,
+            reason="Missing value for required parameter 'title' or 'text_id'.",
+            status=200,
+        )
+
+    db.session.query(TextBookmark).filter(
+        TextBookmark.title == title,
+        TextBookmark.text.has(id=text_id),
+    ).delete()
```
jzohrab, line 86: It's preferable to pass the bookmark's ID for this method and the edit method, rather than the text id and title, which could be ambiguous (unlikely, but possible).

lute/bookmarks/routes.py
```diff
@@ -0,0 +1,114 @@
...
+    new_title = data.get("new_title")
+
+    if not title or not text_id:
+        return jsonify(
+            success=False,
+            reason="Missing value for required parameter 'title' or 'text_id'.",
+            status=200,
+        )
+
+    db.session.query(TextBookmark).filter(
```
jzohrab, line 109: Same here, searching by ID is clearer.

lute/templates/bookmarks/list.html
```diff
@@ -0,0 +1,198 @@
...
+  <thead>
+    <tr>
+      <th style="text-align: left">Page</th>
+      <th style="text-align: left">Description</th>
+      <th style="text-align: left">Actions</th>
+    </tr>
+  </thead>
+</table>
+
+<!-- the config widget for datatables.  Moved into the datatables table footer during setup. -->
```
jzohrab, line 20: I don't think we need this config widget but it doesn't hurt I guess!  Very minor point.

lute/templates/read/index.html
```diff
@@ -529,6 +529,39 @@ <h2>&#127881;</h2>
...
+        "Content-Type": "application/json"
+      },
+      body: JSON.stringify({
+        title: title,
+        book_id: book_id,
+        page_num: page_num,
+      })
+    }).then((resp) => resp.json()).then(json => {
+      if (json?.success === true) {
+        alert("Bookmark added!")
```
jzohrab, line 557: Prefer no exclamation point :-)  -- eg when terms are updated, the flash message is `"{term_text}" updated.`

lute/templates/read/index.html
```diff
@@ -529,6 +529,39 @@ <h2>&#127881;</h2>
...
+      } else {
+        alert(`Unable to add bookmark. ${json?.reason || "Unknown reason"}`)
+      }
+    })
+
+    // TODO: notify user that bookmark was added
```
jzohrab, line 563: Great point, but maybe the alert is sufficient.  Were you thinking of a flash message, or some kind of icon?

jzohrab (changes requested):
This looks super, @cblanken .  My main request is that bookmark edits and deletes should use the bookmark ID to find the database entity.  (Some app designers prefer to use a "business key" rather than expose real entity IDs, but IMO we can stick w/ the IDs.)

If you're too busy to make these changes, LMK.  This PR is very clear (👍🎉😄) and I can make the changes when I test it out locally.  Super stuff!

cblanken:
Thanks for the review. I should have some time to fixup everything tomorrow 👍

lute/templates/read/index.html
cblanken: Yeah, I wasn't sure if you'd want a nicer notification like a flash message or something. I think the `alert` works well enough for now, but I just put it in there as a placeholder.

If the `alert` is good enough for you though, want me to just remove the TODO?

lute/templates/bookmarks/list.html
cblanken: I'd prefer to keep it tbh. Could be useful when I add more stuff to the bookmarks 🙂

cblanken:
Oops. So it begins...me forgetting to run the lint before commiting now that the lint hook is disabled by default :disappointed:

lute/templates/bookmarks/list.html
jzohrab: Sounds good!

jzohrab:
Looks good!  There are a few very small things (e.g. the bookmark title isn't required for a deletion, etc), but I'll fix them up when I check out this branch locally to try it out.  I'll just commit those changes to my local branch, you'll see them when it's merged. I'll let you know if I get stuck, thanks very much @cblanken 👍

jzohrab:
Launched in 3.5.0. :tada:

Issue #391, opened 2024-04-03
Add warning if exiting language edit without saving


jzohrab:
Per discord:

> Anytime within the process of editing or creating a language, the user could click on the Lute icon, on one of the options on the upper right or just close the lute tab by mistake and the entire editing process would be lost. There should, in my opinion, be a small JS message saying "Are you sure you want to leave this page without saving it? If you do all progress will be lost" with the option to either proceed or cancel anytime the user closes the tab or goes to the homepage.

This makes sense, just do a form "dirty" check before leaving, with option to cancel.

jzohrab:
Thank you @cblanken!  Merged into the `develop` branch.

jzohrab:
Launched in 3.5.0. :tada:

PR #444, opened 2024-06-09
Add leave page confirmation dialog to language edit form

cblanken:
This addresses issue #391 to add a confirmation dialog box before navigating away from the language edit form.

It seemed pretty straight-forward so I threw this together. It just adds a "dirty" check like mentioned in the issue and forces a confirmation on the `beforeunload` event if any of the form fields have been changed.

Notably, I had to disable the popup for the save button as well as the delete button so it can use it's own (less generic) confirmation dialogue.

It passed all current tests, but I haven't added any new ones yet. Are there any particular smoke/acceptance tests you'd like me to add @jzohrab?

### DIFF
--- a/lute/templates/language/_form.html
+++ b/lute/templates/language/_form.html
@@ -12,7 +12,7 @@
   {% endfor %}
 {% endfor %}
 
-<form method="POST">
+<form id="edit-lang" method="POST">
   {{ form.hidden_tag() }}
 
   <table id="language">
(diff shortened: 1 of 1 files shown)


jzohrab:
Hi @cblanken - thanks very much!  Re extra testing, let's go with this as-is, since it's so small.  Hopefully that doesn't come back and create problems later :-)  As long as the existing tests pass, we're good to go.

jzohrab:
Launched in 3.5.0. :tada:

Issue #293, opened 2024-02-28
Statistics view gets infinitely small, then loops over.


shetozz:
**Description**

On first navigation to statistics, it gets into a loop of getting smaller then starting over again.

**To Reproduce**

Steps to reproduce the behavior, e.g.:
1. Launch lute
2. Go to About>statistics

**Screenshots**

[Screencast from 2024-02-28 10-01-49.webm](https://github.com/jzohrab/lute-v3/assets/93838524/1a084f9d-71ec-4aac-9362-6cd8f064a479)

**Extra software info, if not already included in the Description:**

- OS: Fedora 39 workstation edition.
- Browser: Brave 1.63.162 (doesn't happen on Firefox)
- How you've installed Lute: pip (using the mandarin fork)
- Version: 3.2.1 (but it's been there for a while)

jzohrab:
Well that's interesting.  Thanks for the issue!

M-Biggles:
What a weird one.

Just going to say I don't have this on the same version of Brave on Linux Mint 21.2.

cblanken:
I don't know why, but this issue seems to only pop up when the browser level zoom is not 100%.

It also works fine for other levels of zoom like 60% and 80% it will shrink slightly and then stop and not loop.
Similarly, 150%, 160%, and some others won't do the infinite loop but a lot of zoom levels in between those will trigger it.

jzohrab:
Ah super thanks @cblanken , I wasn't able to reproduce it, this should help investigation.

cblanken:
@jzohrab I'd like to take on this issue if you haven't looked into yet. It seems like it shouldn't be too complicated.

jzohrab:
Absolutely @cblanken if you have cycles, that would be super!! The js library is in lute/static/js, I believe.

cblanken:
Evidently it's an unfixed issue in ChartJS. See https://github.com/chartjs/Chart.js/issues/10890

One work around is to set the `maintainAspectRatio` option to `false` and then setting the width and height of the canvas container. 

Just as a quick fix, I forced the aspect ratio to 3/1 (so three times as wide as it is tall) and it seems to work pretty well. Here are some screenshots at different resolutions.

## 1920 x 1080
![image](https://github.com/user-attachments/assets/8f4ff2eb-827a-48dd-8af8-de39893a344d)


## 2560 x 1440
[...]

PR #456, opened 2024-08-02
Add workaround for ChartJS shrinking chart bug

cblanken:
Fix for issue #293


jzohrab:
How odd that this fixes it. :-)  Thank you!

PR #461, opened 2024-08-10
Refresh stale bookstats without page reload

cblanken:
Lute's main page fails to reload bookstats from a cached page.

Currently bookstats are only refreshed when a client requests the main page at `/`. When a book is opened, it's bookstats are deleted such that when a user navigates to a cached version of the main page (e.g. they use their browser's back button) the necessary bookstats record is not refreshed.

To remediate this, the `refresh_stats` call has been relocated to a new route at `/refresh_stale_stats` which is called by the root `index.html` to refresh stats whether it has been cached or not.

I've done some limited testing in the dev environment with the default books and this solution seems to work well. If there are any concerns with this approach or new tests that should be added please let me know and I'll implement as soon as I can. Thanks.

### DIFF
--- a/lute/templates/index.html
+++ b/lute/templates/index.html
@@ -33,9 +33,6 @@
   <p>To get started using Lute, first <a href="/language/list_predefined">load a predefined language and sample text</a>, or <a href="/language/new">create your language.</a>
 </p>
 {% else %}
-  <script>
-    fetch("/refresh_stale_stats", { "cache": "no-store" });
-  </script>
   {% include "book/tablelisting.html" %}
 {% endif %}
 
--- a/lute/templates/index.html
+++ b/lute/templates/index.html
@@ -33,6 +33,9 @@
   <p>To get started using Lute, first <a href="/language/list_predefined">load a predefined language and sample text</a>, or <a href="/language/new">create your language.</a>
 </p>
 {% else %}
+  <script>
+    fetch("/refresh_stale_stats", { "cache": "no-store" });
+  </script>
   {% include "book/tablelisting.html" %}
 {% endif %}
 
--- a/lute/app_factory.py
+++ b/lute/app_factory.py
@@ -139,7 +139,6 @@ def _add_base_routes(app, app_config):
         if is_production and have_books and should_run_auto_backup:
             return redirect("/backup/backup", 302)
 
-        refresh_stats()
         warning_msg = backupservice.backup_warning(bkp_settings)
         backup_show_warning = (
             bkp_settings.backup_warn
@@ -162,6 +161,12 @@ def _add_base_routes(app, app_config):
             backup_show_warning=backup_show_warning,
             backup_warning_msg=warning_msg,
         )
+
+    @app.route("/refresh_stale_stats")
+    def refresh_stale_stats():
+        refresh_stats()
+        return {}, 200
+
 
     @app.route("/refresh_all_stats")
     def refresh_all_stats():
--- a/lute/app_factory.py
+++ b/lute/app_factory.py
@@ -167,7 +167,6 @@ def _add_base_routes(app, app_config):
         refresh_stats()
         return {}, 200
 
-
     @app.route("/refresh_all_stats")
     def refresh_all_stats():
         books_to_update = db.session.query(Book).filter(Book.archived == 0).all()
--- a/lute/templates/book/tablelisting.html
+++ b/lute/templates/book/tablelisting.html
@@ -156,13 +156,17 @@
     const current_language_id = {{ current_language_id }};
     // Language filter is set up first, because its value
     // is used during the initial datatable draw.
+
     setup_language_filter(current_language_id);
-    setup_book_datatable("{{ initial_search or '' }}");
 
-    move_datatables_controls_to_config_widget();
+    fetch("/refresh_stale_stats", { "cache": "no-store" })
+    .then(() => {
+      setup_book_datatable("{{ initial_search or '' }}");
+      move_datatables_controls_to_config_widget();
 
-    $("#datatables_config_toggle").click(function() {
-      $("#datatables_config_items").toggle();
+      $("#datatables_config_toggle").click(function() {
+        $("#datatables_config_items").toggle();
+      });
     });
   });
 


cblanken:
A short video of the fix.

https://github.com/user-attachments/assets/9e023501-5335-4ba9-9570-20afc017e11b

lute/templates/index.html
```diff
@@ -33,6 +33,9 @@
   <p>To get started using Lute, first <a href="/language/list_predefined">load a predefined language and sample text</a>, or <a href="/language/new">create your language.</a>
 </p>
 {% else %}
+  <script>
+    fetch("/refresh_stale_stats", { "cache": "no-store" });
+  </script>
```
jzohrab, line 38: This _should_ work reasonably well, but couldn't it run into timing issues, resulting in it still being stale?  The fetch will run causing the stats recalc to start, but then the include will run immediately afterwards and the query might run on data that's still being updated.

The flask `include` that follows calls another endpoint ... I don't think there's an easy way to ensure that this call runs after the `fetch` is completed, but I could be wrong.

jzohrab:
Hey @cblanken - I get the idea but wonder if it will still have staleness issues.  At least it will likely deal with the "back" button problem when you mentioned.  Thoughts?  All in all, I don't have any objections to your idea, it looks like a decent solution to me :-)  Thanks!

lute/templates/index.html
cblanken: Yeah, I was thinking the same thing. It could _technically_ happen. But I think it would be exceedingly unlikely since the fetch should initiate before any of the ajax requests triggered by the Datatable rendered from the following `include`. Worst case, I think the Datatable will just get `null` for the book stat that hasn't be refreshed yet. Which is why the status bar was rendering before.

cblanken: There's probably a more robust way of doing it, but I can't think of it right now.

jzohrab: This feels like an incremental improvement, and even if it's not 100% perfect it's better than the current state.  I'm going to check one thing, we might be able to get the best result by including your fix in the table listing page itself.

cblanken:
I don't really think there any problems with the "staleness" mechanism. As far as I can tell, an opened book's stats are always deleted when opening so they should be refreshed when the main page is re-opened. The main issue is the caching. Actually thinking about it some more, it may be better to just restrict caching for the main page with an HTTP header (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control). It would basically force a reload on the main page whenever you navigated to it but I think it would fix it without the data-race concerns.

lute/templates/index.html
jzohrab: `templates/book/tablelisting.html` has the following:

```
  $(document).ready(function () {
    ...
    setup_book_datatable("{{ initial_search or '' }}");
    ...
```

If we move that fetch to above the "setup" method, and have `setup_book_datatable` get called on success, then it should resolve any timing issues.  But that could be another issue, as your fix here seems to be a good initial step.  Sounds good, @cblanken ?

cblanken: I can go ahead an fix it up. No need to make it a separate issue.
Were you thinking like this? There isn't really any error checking on the `refresh_stale_stats` endpoint. It just executes the refresh and returns.

```js
  $(document).ready(function () {
    const current_language_id = {{ current_language_id }};
    // Language filter is set up first, because its value
    // is used during the initial datatable draw.

    setup_language_filter(current_language_id);

    fetch("/refresh_stale_stats", { "cache": "no-store" })
    .then(() => {
      setup_book_datatable("{{ initial_search or '' }}");
      move_datatables_controls_to_config_widget();

      $("#datatables_config_toggle").click(function() {
        $("#datatables_config_items").toggle();
      });
    });
  });
```

That should force the execution to be synchronous and it seems to work fine as far as I can tell.

jzohrab: Yep just like that. :-)

jzohrab:
> I think it would fix it without the data-race concerns.

Ah right maybe that's easiest, if it works.  I'm outside of my realm of limited expertise here.

cblanken:
> > I think it would fix it without the data-race concerns.
> 
> Ah right maybe that's easiest, if it works. I'm outside of my realm of limited expertise here.

A bit out of my wheelhouse too to be honest.

The synchronous fetch accomplishes the same thing though. And this way there's no need to rely on the user's browser to respect the caching headers. Although I'm sure they would.

cblanken:
I don't know why all those tests failed, I'll have to look into them later.

jzohrab:
> I don't know why all those tests failed, I'll have to look into them later.

Re-ran them manually to check, still failing.  Those are acceptance tests failures, run with `inv accept`.  (stop your current running dev lute before running it :-) )

cblanken:
@jzohrab Okay, so I'm not super familiar with how pytest-bdd works, but I think they're all failing because the tests don't await for the table load. Adding a short `_sleep` to the `given_visit` let's all the acceptance tests pass.

```python
@given(parsers.parse('I visit "{p}"'))
def given_visit(luteclient, p):
    "Go to a page."
    luteclient.visit(p)
    _sleep(0.1)
```

I really don't like that solution, but I don't have time to really dig into refactoring all those failing test to await the table load with a timeout instead of using `_sleep`s.

jzohrab:
Adding a sleep is fine, it's a minimal delay.  I'll check the final failure.

jzohrab:
Ah I guess you need to push that commit that adds the delay :wave: -- ping me when it's pushed, @cblanken .  Thank you again!

cblanken:
@jzohrab Sorry, I was mistaken. There appears to be some other issues even with the extra `sleep`. The selenium/splinter errors are too opaque for me to figure what's going wrong though.

Moving the DataTable initialization out of the `.then` seems to let all the tests pass though it throws some SQLAlchemy errors into the output too.

I can revert it to my original solution if you want, but trying to pick apart the tests is too much for me at the moment.

jzohrab:
Cheers @cblanken I’ll check it out. Thanks!

jzohrab:
Looked briefly, but thought that your idea of not caching the main page might be better, so am trying that out in #468.  Let's see how the tests do.

jzohrab:
@cblanken - the non-caching approach of #468 worked, all tests pass.  Seems like a reasonable approach to me -- great idea! -- so I'll merge that in and will close this one.  Cheers!

PR #468, opened 2024-08-15
Never cache main page

jzohrab:
Disable main page cache to always do pending book stat recalculations.

Seeing if this is a simpler approach than #461, which tries to do the same thing with javascript, but runs into some acceptance test problems.


jzohrab:
Launched in 3.5.1.

Issue #469, opened 2024-08-16
3.5.1: Dictionaries no longer loading


jamesdeluk:
Upgraded to 3.5.1, dictionaries typically don't load. Perhaps 1 time in 10 they load.

Downgraded to 3.5.0, they load again every time.

Specifically, https://korean.dict.naver.com/koendict/#/search?query=### and https://en.dict.naver.com/#/search?query=###

Possibly related to https://github.com/LuteOrg/lute-v3/issues/464 and https://github.com/LuteOrg/lute-v3/pull/459?

jzohrab:
Well that's a drag, sorry about that!  I tested with some other dicts and it was fine, but obv. it's no good.  So I'll revert that change and maybe we can work something else out.  Should be out soon, thanks.

jamesdeluk:
It happens 😄 Maybe someone else can double-check to ensure it's not just an issue I'm having.

cblanken:
> It happens 😄 Maybe someone else can double-check to ensure it's not just an issue I'm having.

@jamesdeluk I'm having trouble replicating this. Could you post what browser you're using and the version?

For me they load immediately, but it's possible they won't display immediately if your connection is slow. Do they show up if you wait a while (10-15 seconds) after selecting a term or does the dictionary pane just keep the background color without any indication of loading?

Also if you have any other dictionaries failing to load please post them as well so I can check them. Thanks.

jamesdeluk:
Firefox 129.0.2 (64-bit) on Windows 11.

I've just re-upgraded to 3.5.1 and the issue persists.

Waiting for extended periods doesn't work.

Other dictionaries (e.g. https://en.wiktionary.org/wiki/###) load fine.

The network does do a GET request for the URL, but nothing loads.

When I click off and on again, I get `en.dict.naver.com content already loaded.` (or `korean.dict.naver.com content already loaded.`) in the console, so it thinks it's loaded.

Downgrading to 3.5.0, everything loads again instantly. After the GET request, there are a bunch more as the page loads.

jzohrab:
Hey @jamesdeluk , I'm going to back out the change in 3.5.1.  Too bad :-/  but it's causing other issues too.  @cblanken spent a whack of time trying to come up with workarounds, but no dice.  Thanks for the detail!

jzohrab:
Change reverted, this should be fixed when 3.5.2 is launched.  Thanks @jamesdeluk and @cblanken , cheers!

[...]

PR #473, opened 2024-08-24
Revert dict tab focus fix and sandbox dict tab iframes

cblanken:
One possible solution to prevent dictionary iframes from stealing focus is to
enable the CSP sandbox on each iframe. If a bare `sandbox` attribute is used,
then autofocus from iframes should be blocked. Unfortunately this has the side
effect of blocking some online dictionary that require additional feature
policies such as `allow-same-origin` and `allow-scripts`.

In particular the following dictionaries mentioned in issue https://github.com/LuteOrg/lute-v3/issues/469 require these policies
to load correctly:
- https://korean.dict.naver.com/koendict/#/search?query=###
- https://en.dict.naver.com/#/search?query=###

### DIFF
--- a/lute/static/css/styles.css
+++ b/lute/static/css/styles.css
@@ -308,10 +308,6 @@ div#read_pane_left {
     width: 50%;
 }
 
-#read_pane_left:focus {
-    outline: none;
-}
-
 div#read_pane_right {
     display: grid;
     grid-template-rows: 18rem 1fr;
--- a/lute/static/js/dict-tabs.js
+++ b/lute/static/js/dict-tabs.js
@@ -29,6 +29,7 @@ class LookupButton {
     this.contentLoaded = false;
 
     this.frame = createIFrame(frameName);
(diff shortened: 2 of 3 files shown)


jzohrab:
Hi @cblanken - I wonder if this should be split out further, with one PR just to revert the existing change, and another to add the sandbox etc attributes in that new line.  I wouldn't want to "fix" one problem just to cause another.  Then we could look at the extra policy stuff more carefully.  Thoughts?

jzohrab:
Chatted briefly w/ @cblanken on discord, this is a draft idea to investigate sandboxing only.  I'll revert the old PR, will keep this as an idea only for sandboxing investigations.  Thanks!!

jzohrab:
Thanks for the PR!  I'm closing it as it is either stale, or not ready to merge in its current form. I've added the "future-work" label because there are ideas and implementation details here that could be useful if someone revisits this feature in the future.

Issue #324, opened 2024-03-13
Auto-convert pending parent tag to real tag on Save


bill-hkc:
(edited from the original issue)

**Description**

I try to add a parent term.  I could add it successfully when the parent term was defined.  ~~When the parent term wasn't defined previously, it is not automatically created.   Thus I could not add it as the parent term.~~

When the parent term doesn't exist, it's not clear that you have to hit Return to convert it to a tag, which is then saved.  What should happen: any pending text in the "parents" text box should be auto-converted to a tag when the form is saved.

**To Reproduce**

Steps to reproduce the behavior, e.g.:

1. open a book for reading
2. click on a term to open the term window, ex "entitled"
3. fill the parents field, ex "entitle"
4. click save button.  The term window disappear with a message "entitled updated".
5. click on the term "entitled" to open the term window again, and find the parents field still blank.

**Screenshots**
![image](https://github.com/jzohrab/lute-v3/assets/7263899/2bdb0eda-a804-44b3-b8d2-51641568e8c2)

**Extra software info, if not already included in the Description:**

- windows 10
- chrome
- installation: python pip
- Version:  3.2.4

jzohrab:
Dev notes:

- the parent tags are managed by tagify (https://github.com/yairEO/tagify).  Per the repo readme, tagify has a "addTagOnBlur" property, but we **can't** use that for Lute as sometimes external dictionaries pull focus mid-parent-tag typing, which causes problems
- should add some kind of handler to the Javascript save to convert any pending parent input text to a fully-fledged tag.  I'm not sure exactly what that would entail, perhaps doing a "sendkeys" or something to simulate a user "enter" in the tag textbox control.

jzohrab:
I tried to add an "onsubmit" event for the form to convert any outstanding tag text to real tags, but they weren't getting submitted properly.  I'm not sure why, so I opened a question in the tagify repo.  https://github.com/yairEO/tagify/issues/1375 -- hopefully something will turn up.

jzohrab:
Fixed in `develop`.

jzohrab:
Launched in `3.5.3`, should be fixed now, thank you!

PR #477, opened 2024-09-09
Issue 324 pending parent tag

jzohrab:
Addresses #324


jzohrab:
Local tests are failing, trying here to check.

PR #491, opened 2024-10-08
Add CLI tool to bulk import books from a CSV file.

bwkimmel:
This adds a command-line tool to bulk-import books from a CSV file.

The CSV file must have have a header row and contain specific columns which map to the properties of a book. For details, see the help for the new command:

```
flask --app lute.app_factory cli import_books_from_csv --help
```

### DIFF
--- a/lute/book/model.py
+++ b/lute/book/model.py
@@ -50,6 +50,13 @@ class Repository:
             raise ValueError(f"No book with id {book_id} found")
         return self._build_business_book(dbb)
 
+    def find_by_title(self, book_title):
+        "Loads a Book business object for the book with a given title."
+        dbb = DBBook.find_by_title(book_title)
+        if dbb is None:
+            return None
+        return self._build_business_book(dbb)
+
     def get_book_tags(self):
         "Get all available book tags, helper method."
         bts = self.db.session.query(BookTag).all()
--- a/lute/book/model.py
+++ b/lute/book/model.py
@@ -50,9 +50,9 @@ class Repository:
             raise ValueError(f"No book with id {book_id} found")
         return self._build_business_book(dbb)
 
-    def find_by_title(self, book_title):
+    def find_by_title(self, book_title, language_id):
         "Loads a Book business object for the book with a given title."
-        dbb = DBBook.find_by_title(book_title)
+        dbb = DBBook.find_by_title(book_title, language_id)
         if dbb is None:
             return None
         return self._build_business_book(dbb)
--- a/lute/cli/commands.py
+++ b/lute/cli/commands.py
@@ -6,6 +6,7 @@ import click
 from flask import Blueprint
 
 from lute.cli.language_term_export import generate_language_file, generate_book_file
+from lute.cli.import_books import import_books_from_csv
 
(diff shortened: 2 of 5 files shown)


lute/book/model.py
```diff
@@ -50,6 +50,13 @@ def load(self, book_id):
             raise ValueError(f"No book with id {book_id} found")
         return self._build_business_book(dbb)
 
+    def find_by_title(self, book_title):
```
jzohrab, line 53: This should probably also include the book language as a parameter, e.g if I'm learning Spanish and French, and both have a book called "lesson_1".  Small point only :-)

jzohrab:
Nice stuff!  Great help text for that command, thank you very much, such things are very helpful. :-)

I think that the Language name should be added as a parameter to the method -- but maybe that's overthinking it, it's a first pass at the code.  What are your thoughts?

I think this is a good start.  I'd like to expand this so that users could pass in text file names, but currently the code in Lute isn't that great for such things -- iirc the text file parsing is buried somewhere in the forms.

jzohrab:
Just a few lint errors, @bwkimmel :-).   There's a task to check that locally as well: `inv lint`

lute/book/model.py
bwkimmel: Done.

bwkimmel:
Those changes sound good to me! I've gone ahead and made that change, and also:
  - Added the option to include the language in the CSV file itself (making the `language` command-line parameter optional).
  - Added help text for the command-line options.
  - Included audio file and bookmarks in the unit test (since you had indicated to me that the contents of the audio file isn't actually checked, so a dummy file would suffice).
  - Fixed the linter errors

Having the option to provide the text in the CSV file or as a reference to an external file (like the way audio files are provided) sounds good to me!

I've left the changes as separate commits so that you can see what they are. Let me know if you want them squashed before merging, though!

jzohrab:
Thanks @bwkimmel I'll check this out soon, appreciate the fast changes.  No worries on squashing, overall it's a well-contained change and everything gets merged no-ff anyway.  Cheers!

jzohrab:
ps @bwkimmel -- don't worry about the "black" reformat failing, I'm going to check this out locally first anyway, and will merge manually.  I'll fix that at that point.

(The "black" code check is a style check, it's handled in one of the pre-commit checks, or with `inv black` :-P )

jzohrab:
I haven't tried it out myself yet, but this seems like a good step forward, and provides a good building base for future stuff.  Thank you!  I'll merge this into the dev branch now.

Great help content again, thanks.  We should add this to the manual too, https://luteorg.github.io/lute-manual/usage/command-line-jobs.html.  @bwkimmel , could you add a PR to the manual repo too with doc notes?  Just the help content is good, and maybe a sample one-line CSV file and sample CLI call, like the other things.  I'll open an issue over there: https://github.com/LuteOrg/lute-manual/issues/5 -- if you just want to dump some comments in the issue, I can make the changes to the manual as I have it checked out and the tools set up on my machine.

jzohrab:
Launched in 3.5.6.

PR #510, opened 2024-11-02
Jdom/plugin/thai

jaydom28:


plugins/lute-thai/lute_thai_parser/parser.py
```diff
@@ -0,0 +1,69 @@
...
+        text = text.replace("\r\n", "\n")
+
+        words = pythainlp.word_tokenize(text)
+        tokens = []
+        pattern = f"[{language.word_characters}]"
+        for word in words:
+            is_word_char = re.match(pattern, word) is not None
+            is_end_of_sentence = word in language.regexp_split_sentences
+            if is_end_of_sentence:
+                is_word_char = False
```
jaydom28, line 54: Just wanted to leave a comment about this logic here,

In Thai the "period" punctuation is `ฯ` and in my tests, the parser was showing it as a word and I figured it didn't make sense for punctuation to be words so that is why I have `is_word_char` set to `False` if the text is a sentence delimiter

jzohrab: Right, sometimes there maybe funny cases like this where you have to hardcode it.

jzohrab:
Looks great, thanks!

jzohrab:
Released to pypi woot.

PR #513, opened 2024-11-09
added khmer parser plugin

jaydom28:


plugins/lute-khmer/lute_khmer_parser/parser.py
```diff
@@ -0,0 +1,76 @@
...
+            is_end_of_sentence = word in language.regexp_split_sentences
+            is_whitespace = re.match(whitespace_regex, word) is not None
+            if is_whitespace:
+                continue
+
+            is_word_char = (not is_end_of_sentence) and (
+                re.match(pattern, word) is not None
+            )
+
+            if word == "NEWLINE_CHARACTER_FOR_LUTE":
```
jaydom28, line 63: When a text is parsed using `khmernltk` the `\n` characters just get erased. Because of this, I tried replacing newlines with something with a phrase and compared against that phrase instead

lute/db/language_defs
jaydom28, line 1: This is an accident, I am trying to figure out how to remove this

plugins/lute-khmer/lute_khmer_parser/parser.py
```diff
@@ -0,0 +1,86 @@
...
+                is_end_of_sentence = True
+
+            if word.startswith("\\"):
+                num_leading_slashes = len(word) - len(word.lstrip("\\"))
+                for _ in range(num_leading_slashes):
+                    tokens.append(ParsedToken("¶", False, True))
+
+                word = word.lstrip("\\")
+                is_word_char = True
+                is_end_of_sentence = False
```
jaydom28, line 76: Note about this logic here, the `khmernltk` parser completely gets rid of `\n` characters. What I decided to do instead is replace all the `\n` with `\\`, resulting in some words coming out as `\\ខ្ញុំ`. This logic replace any preceding slashes with the special character and then parses the original token like normal

plugins/lute-khmer/lute_khmer_parser/parser.py
jzohrab: LGTM, cheers.  This can be worked on as things evolve, if needed.

lute/db/language_defs
jzohrab: Thanks, resolved.  That language_defs submodule is a bit of a hassle, but good enough for now.

plugins/lute-khmer/lute_khmer_parser/parser.py
jzohrab: Good approach if it's working for you :-)

jzohrab:
In 3.6.0 release.

PR #538, opened 2024-12-16
Black and white theme for E-Ink Displays

Christhomas17:
A simple theme for a black and white E-ink display. Underlines are used to distinguish level 1-5 words from unknown words. The color scale use for level 1-3 is a darker to lighter grey.

### DIFF
--- a/lute/themes/css/Black & White Display.css
+++ b/lute/themes/css/Black & White Display.css
@@ -1,0 +1,65 @@
+body {
+  --status-0-color: #bdbdbd;
+  --status-1-color: #bdbdbd;
+  --status-2-color: #e0e0e0;
+  --status-3-color: #f5f5f5;
+  --status-4-color: transparent;
+  --status-5-color: transparent;
+
+  --status-98-color: transparent;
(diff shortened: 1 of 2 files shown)


jzohrab (changes requested):
As discussed in discord, the colors are a bit dark for most users.  Try tweaking them, see what you can get to.

If you're familiar with git rebase, maybe after your changes squash it to a single commit, and use this commit message:

```
Add simple theme for black and white E-ink display.
    
Underlines are used to distinguish level 1-5 words from unknown words.
The color scale use for level 1-3 is a darker to lighter grey.
```

If not, no worries, I'll fix the message.

jzohrab:
Hi @Christhomas17 , thanks for the update.  Do you think these are ready to go?

I don't think an ampersand in the filename is a good idea, best to stick with alphanumeric.  Maybe `Black_and_white.css` (also avoiding spaces in the filename is good, just me being simpleminded).

jzohrab:
Closing, I assume this is superceded by #557 .  Thanks!

PR #568, opened 2025-01-10
Add Mobile Focus Mode

Christhomas17:
To alleviate the time it takes while on mobile to set unknown words to status = 1 and type a definition, a new mobile focus mode was added. While mobile focus mode is enabled, when an unknown word is clicked, the status is set to 1 and no popups are shown.

### DIFF
--- a/lute/static/js/lute.js
+++ b/lute/static/js/lute.js
@@ -82,6 +82,8 @@ const _isUserUsingMobile = () => {
     return false;
   if (s == 'mobile')
     return true;
+  if (s == 'mobile-focus')
+    return true;
 
   // User agent string method
   let isMobile = /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent);
@@ -109,6 +111,16 @@ const _isUserUsingMobile = () => {
   }
 
   return isMobile
+
+}
+
+//Determine if Mobile focus has been turned on
+const _is_mobile_focus = () => {
+  const s = localStorage.getItem('screen_interactions_type');
+  if (s == 'mobile-focus')
+    return true
+  else
+    return false
 }
 
 
@@ -572,7 +584,14 @@ function _single_tap(el, e) {
   clear_newmultiterm_elements();
   const term_is_status_0 = (el.data("status-class") == "status0");
   if (term_is_status_0) {
-    show_term_edit_form(el);
+    if(_is_mobile_focus){
+      el.addClass('kwordmarked')
+      el.addClass('wordhover')
+      update_status_for_marked_elements(1)
+    }  
+    else{
+      show_term_edit_form(el);
+    }    
   }
 }
 
--- a/lute/templates/read/reading_menu.html
+++ b/lute/templates/read/reading_menu.html
@@ -124,6 +124,11 @@
           </a>
         </li>
         <li>
+          <a id="screen-mobile-focus" class="reading-menu-item" href="" onclick="set_screen_type('mobile-focus'); return false;">
+            Mobile Focus
+          </a>
+        </li>
+        <li>
           <a id="screen-desktop" class="reading-menu-item" href="" onclick="set_screen_type('desktop'); return false;">
             Desktop
           </a>


jzohrab:
Thanks @Christhomas17 - I'll try to get to this shortly.  Sorry for radio silence, things have been tight lately.

lute/static/js/lute.js
```diff
@@ -109,6 +111,16 @@ const _isUserUsingMobile = () => {
   }
 
   return isMobile
+
+}
+
+//Determine if Mobile focus has been turned on
+const _is_mobile_focus = () => {
+  const s = localStorage.getItem('screen_interactions_type');
+  if (s == 'mobile-focus')
```
jzohrab, line 120: Can just be `return (s == 'mobile-focus');`

jzohrab: You're defining a function call here, intended?

lute/static/js/lute.js
```diff
@@ -572,7 +584,14 @@ function _single_tap(el, e) {
   clear_newmultiterm_elements();
   const term_is_status_0 = (el.data("status-class") == "status0");
   if (term_is_status_0) {
-    show_term_edit_form(el);
+    if(_is_mobile_focus){
```
jzohrab, line 587: _is_mobile_focus is a function, not a constant

lute/static/js/lute.js
```diff
@@ -572,7 +584,14 @@ function _single_tap(el, e) {
...
+      el.addClass('kwordmarked')
+      el.addClass('wordhover')
+      update_status_for_marked_elements(1)
+    }  
+    else{
+      show_term_edit_form(el);
+    }
```
jzohrab, line 594: Lots of nested ifs get hard to follow, could be written as:

```
  if (el.data("status-class") != "status0")
    return;
  if (_is_mobile_focus()) { ... etc }
  else { ... }
```

Acknowledging that this might be a personal preference :-)

lute/static/js/lute.js
```diff
@@ -572,7 +584,14 @@ function _single_tap(el, e) {
   clear_newmultiterm_elements();
   const term_is_status_0 = (el.data("status-class") == "status0");
   if (term_is_status_0) {
-    show_term_edit_form(el);
+    if(_is_mobile_focus){
+      el.addClass('kwordmarked')
+      el.addClass('wordhover')
```
jzohrab, line 589: only need one style, not both

lute/templates/read/reading_menu.html
```diff
@@ -123,6 +123,11 @@
             Mobile
           </a>
         </li>
+        <li>
+          <a id="screen-mobile-focus" class="reading-menu-item" href="" onclick="set_screen_type('mobile-focus'); return false;">
```
jzohrab, line 127: I'm not sure about this -- the screen type is still `mobile`, you just happen to be changing something about how the interactions are done on mobile.  It feels like this should set some different kind of local storage setting.

lute/static/js/lute.js
```diff
@@ -82,6 +82,8 @@ const _isUserUsingMobile = () => {
     return false;
   if (s == 'mobile')
     return true;
+  if (s == 'mobile-focus')
```
jzohrab, line 85: As mentioned later, I'm not sure about "mobile-focus" being a "screen type", if that makes sense.  This might feel nitpicky, I'm not sure.  Just feels iffy to me at first glance.

lute/templates/read/reading_menu.html
```diff
@@ -123,6 +123,11 @@
...
+            Mobile Focus
```
jzohrab, line 128: People will need some kind of documentation on what this means.  There is a "desktop focus", but this is different, you're changing how clicks work.  Not sure what the right place for this is, but I don't think this is it.  Hard to say.

Edit: on the other hand, this _is_ defining/refining "screen interactions".  Still feels off, and I can't explain why.

> image

jzohrab (changes requested):
Some code changes, but mostly questions about overall UX, and where this feature really belongs.  It's not a new screen type ... hm, it feels like it's shoehorned in to get it working, but that might get confusing later.  Unfortunately I don't have a clearer response to you at the moment.

jzohrab:
After some discussion in Discord, this PR is replaced by #569 .  :wave:  Thanks @Christhomas17 for opening it.

Issue #563, opened 2025-01-08
Don't block image search page if bing is slow


jzohrab:
Currently, the image search makes a request to bing and waits for a response before showing the image list.  That can look slow or weird if bing is having trouble.

Instead the page should load instantly, and then make a separate ajax request to get the bing image search content.

jzohrab:
"good first issue" as I think this is relatively self-contained.

parradam:
Hey @jzohrab ! I'm obviously very new to the codebase, but thought I could have a crack at this one.

If I understand you correctly, the UI thread is blocked while waiting for the image search response from Bing.

I would propose something like the code below. Essentially showing a loading message and an then making an async `fetch` for the Bing content, then displaying it. Seems to work as intended.

Please let me know if this is what you were after - happy to discuss further and raise a PR!

PS: I see that we could probably cache the images too, maybe that's a further improvement!

```js
class ImageLookupButton extends GeneralLookupButton {
  constructor() {

    // Parents are in the tagify-managed #parentslist input box.
    let _get_parent_tag_values = function() {
      const pdata = LookupButton.TERM_FORM_CONTAINER.querySelector("#parentslist").value
      if ((pdata ?? '') == '')
        return [];
      return JSON.parse(pdata).map(e => e.value);
    };

    let handler = async function(iframe) {
      const text = LookupButton.TERM_FORM_CONTAINER.querySelector("#text").value;
      const lang_id = LookupButton.LANG_ID;
      if (lang_id == null || lang_id == '' || parseInt(lang_id) == 0 || text == null || text == '') {
        alert('Please select a language and enter the term.');
        return;
      }
      let use_text = text;

      // If there is a single parent, use that as the basis of the lookup.
      const parents = _get_parent_tag_values();
      if (parents.length == 1)
        use_text = parents[0];

      const raw_bing_url = 'https://www.bing.com/images/search?q=[LUTE]&form=HDRSC2&first=1&tsc=ImageHoverTitle';
      const binghash = raw_bing_url.replace('https://www.bing.com/images/search?', '');
      const url = `/bing/search/${LookupButton.LANG_ID}/${encodeURIComponent(use_text)}/${encodeURIComponent(binghash)}`;

      // Changed code
      iframe.srcdoc = '<p>Loading images from Bing...</p>';

      try {
        const response = await fetch(url);
        if (!response.ok) {
          throw new Error(`Error: ${response.statusText}`);
        }

        const data = await response.text()
[...]

PR #570, opened 2025-01-16
Make Bing search asynchronous

parradam:
This PR removes the template tags for images, and replaces them with a JS call once the DOM has loaded. Relates to #563 

Just wanted to push at this stage as:

1. This seems to work some of the time, but not always. I think I've narrowed it down to the Flask route `bing_search` because it seems like the images sources aren't always passed into the template
2. Would it be better if I looked at moving all of the parsing logic into the JS? Essentially porting the code below to JS. Maybe that's what you meant :-)

```python
    # dump("searching for " + text + " in " + language.getLgName())
    search = urllib.parse.quote(text)
    params = searchstring.replace("[LUTE]", search)
    params = params.replace("###", search)  # TODO remove_old_###_placeholder: remove
    url = "https://www.bing.com/images/search?" + params
    content = ""
    error_msg = ""
    try:
        with urllib.request.urlopen(url) as s:
            content = s.read().decode("utf-8")
    except urllib.error.URLError as e:
        content = ""
        error_msg = e.reason

    # Sample data returned by bing image search:
    # <img class="mimg vimgld" ... data-src="https:// ...">
    # or
    # <img class="mimg rms_img" ... src="https://tse4.mm.bing ..." >

    def is_search_img(img):
        return not ('src="/' in img) and ("rms_img" in img or "vimgld" in img)

    def build_struct(image):
        src = "missing"
        normalized_source = image.replace("data-src=", "src=")
        m = re.search(r'src="(.*?)"', normalized_source)
        if m:
            src = m.group(1)
        return {"html": image, "src": src}

    raw_images = list(re.findall(r"(<img .*?>)", content, re.I))

    images = [build_struct(i) for i in raw_images if is_search_img(i)]

    # Reduce image load count so we don't kill subpage loading.
    # Also bing seems to throttle images if the count is higher (??).
    images = images[:25]
```

### DIFF
--- a/lute/bing/routes.py
+++ b/lute/bing/routes.py
@@ -5,10 +5,41 @@ Getting and saving bing image search results.
 import os
 import re
 import urllib.request
-from flask import Blueprint, request, Response, render_template, jsonify, current_app
+from flask import (
+    Blueprint,
+    request,
+    Response,
+    render_template,
+    jsonify,
+    current_app,
+    url_for,
+)
 
 
 bp = Blueprint("bing", __name__, url_prefix="/bing")
+
+
+@bp.route(
+    "/search_page/<int:langid>/<string:text>/<string:searchstring>", methods=["GET"]
+)
+def bing_search_page(langid, text, searchstring):
+    """
+    Load initial empty search page, passing real URL for subsequent ajax call to get images.
+
+    Sometimes Bing image searches block or fail, so providing the initial empty search page
+    lets the user know work is in progress.  The user can therefore interact with the page
+    immediately. The template for this route then makes an ajax call to the "bing_search()"
+    method below which actually does the search.
+    """
+
+    # Create URL for bing_search and pass into template.
+    search_url = url_for(
+        "bing.bing_search", langid=langid, text=text, searchstring=searchstring
+    )
+
+    return render_template(
+        "imagesearch/index.html", langid=langid, text=text, search_url=search_url
+    )
 
 
 @bp.route("/search/<int:langid>/<string:text>/<string:searchstring>", methods=["GET"])
@@ -60,12 +91,13 @@ def bing_search(langid, text, searchstring):
     # Also bing seems to throttle images if the count is higher (??).
     images = images[:25]
 
-    return render_template(
-        "imagesearch/index.html",
-        langid=langid,
-        text=text,
-        images=images,
-        error_message=error_msg,
+    return jsonify(
+        {
+            "langid": langid,
+            "text": text,
+            "images": images,
+            "error_message": error_msg,
+        }
     )
 
 
--- a/.gitignore
+++ b/.gitignore
@@ -117,6 +117,9 @@ ENV/
 env.bak/
 venv.bak/
 
+# VS Code config
+.vscode/
+
 # Spyder project settings
 .spyderproject
 .spyproject
--- a/lute/static/js/dict-tabs.js
+++ b/lute/static/js/dict-tabs.js
@@ -132,7 +132,7 @@ class ImageLookupButton extends GeneralLookupButton {
 
       const raw_bing_url = 'https://www.bing.com/images/search?q=[LUTE]&form=HDRSC2&first=1&tsc=ImageHoverTitle';
       const binghash = raw_bing_url.replace('https://www.bing.com/images/search?', '');
-      const url = `/bing/search/${LookupButton.LANG_ID}/${encodeURIComponent(use_text)}/${encodeURIComponent(binghash)}`;
+      const url = `/bing/search_page/${LookupButton.LANG_ID}/${encodeURIComponent(use_text)}/${encodeURIComponent(binghash)}`;
 
       iframe.setAttribute("src", url);
     };  // end handler
--- a/lute/templates/imagesearch/index.html
+++ b/lute/templates/imagesearch/index.html
@@ -26,25 +26,10 @@
         then paste from your clipboard.
       </p>
 
-      {% for image in images %}
-      <span class="initial"
-            style="margin: 2px; display:inline-block;"
-            onmouseover="highlight_image(this);"
-            onmouseout="un_highlight_image(this);"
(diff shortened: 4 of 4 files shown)


jzohrab:
Hey @parradam - thanks for giving this one a shot :-) appreciated.

Re point 2, moving it into JS: I wasn't thinking of that, b/c the python code to parse the returned results and get the real image sources works, and it _ought_ to be relatively easy to still call that method from ajax, but modify it so that it returns just a list of image URLs.  I don't think that moving that call and the parsing to JS reduces any complexity, it just moves code and rewrites it.  So, I feel that something like this should work:

```
# ... everything the same in the above, up to
images = images[:25]
return jsonify(images)
```

and that struct would then be handled by the ajax call from the template.  Does that make sense?

Re point 1: "This seems to work some of the time, but not always." -- sometimes I get weird blocky things in my image list, and I haven't bothered to look into it :-)

👋

jzohrab:
(assigning to you b/c you're still hacking at it, just so I see it on the project board)

parradam:
Hey @jzohrab ! Happy to keep hacking away :-) hopefully with more sleep behind me this time!

Thanks for your patience and for your input! So here is my thought process, based on our conversation so far, just to ensure we are on the same page. The `bing_search` route does two things at the moment:
1. It collects the image `src`s from Bing
2. Then it renders the `imagesearch/index.html` template, passing them in

- [ ] To decouple these, I presume we'll need a separate route that could be called something like `image_search_page` with a path of `/search_page/<int:langid>/<string:text>/<string:searchstring>`, which accepts the existing URL params and renders `imagesearch/index.html`, passing in the URL that will be needed to call Bing once the DOM has loaded
- [ ] Then we need to update `imagesearch/index.html` to implement the AJAX call using the URL passed in
- [ ] Update `bing_search` to render the jsonified image data as you suggested - everything else in this route remains pretty much the same

On the plus side, it looks like Bing is talking to me again. I did some debugging yesterday on that, and will see if there's anything I can do with the headers (such as `User-Agent`) to prevent an empty `<body></body>` being returned.

Hope this is useful, and if it sounds sensible, I'm happy to get this done tomorrow!

jzohrab:
Hey @parradam - that sounds like what I was thinking too.  Thx for chatting about it to clarify -- writing a good spec takes time and sometimes I skip it :-)  Cheers!

parradam:
Hey @jzohrab , no problem :-)

I've force pushed an updated approach - see what you think! Usual issues with Bing not playing ball, but I was able to experiment returning an object with an `src` property from Flask.

Any thoughts on testing? I guess we could mock the Bing response and check the image renders, might be a pain though?

jzohrab:
Will take a look soon, am sick. 🤒 

Yeah testing is a big hassle and I didn’t bother :-). During testing there’s a variable set to disable images actually.

lute/bing/routes.py
```diff
@@ -5,12 +5,36 @@
...
 
 
 bp = Blueprint("bing", __name__, url_prefix="/bing")
 
 
+@bp.route(
+    "/search_page/<int:langid>/<string:text>/<string:searchstring>", methods=["GET"]
+)
+def bing_search_page(langid, text, searchstring):
+    "Load search page."
```
jzohrab, line 26: Could you add a bit more detail explaining why this is done this way?  Here's my first draft, see how it feels, tweak as needed. :-)

```
"""
Load initial empty search page, passing real URL for subsequent ajax call to get images.

Sometimes Bing image searches block or fail, so providing the initial empty
search page lets the user know work is in progress.  The template for this route
then makes an ajax call to the "bing_search()" method below which actually does the
search.
"""
```

jzohrab (changes requested):
Hi @parradam - Haven't tried the code but it looks great!  One small comment tweak requested.  Is this still draft, or do you think it's good to go?

jzohrab:
(PRs go to the dev branch, master is for launched code)

parradam:
@jzohrab Hope you're feeling better!

Appreciate the feedback! Have incorporated your comment, I'd just like to make sure that Bing behaves as expected before I mark it as ready. Should be able to confirm that in the next day or two.

Thanks for fixing the base branch, I realised and fixed it locally but didn't push it - oops!

parradam:
@jzohrab quick update! I've been experimenting with this and I get a blank page back which is quite annoying:

```html
<!DOCTYPE html><html xml:lang="en" xmlns=http://www.w3.org/1999/xhtml xmlns:Web=http://schemas.live.com/Web><head><meta http-equiv="Content-Type" content="text/html; charset=UTF-8" /><title>Bing</title><meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" /><meta name="viewport" content="width=device-width, initial-scale=1"></head><body></body></html>
```

I've been messing around with different headers (user agent, referer, accept-language, accept) but no luck so far. I can return a dummy image by subbing in `images = [{"src": "url.com/image_location"}] and it behaves as expected. Wonder if I'm blacklisted forever!

Thought I'd raise a flag now before I go too far down this rabbit hole!

jzohrab:
Thanks @parradam , I’ll try it out today. I’ve had times when the URL went dark, maybe periodic blacklisting. Clears up eventually, so far. Will let you know, thanks!

jzohrab:
@parradam - it's working fine for me, are you still seeing a blank screen?  Looks great!

> image

parradam:
@jzohrab amazing! Yes, still dark for me. Possibly due to the amount of experimentation I did! Thanks so much for testing it!

I'll just push my commit (from memory it was just the docstring but will check) and then I'll mark it for review. :-)

parradam:
I've rebased onto Lute's `development` branch and force pushed.

Double checked and all tests seem to be passing.

Hope it's ok! :-)

lute/bing/routes.py
parradam: Hope this is sorted now!

jzohrab (approved):
LGTM, I'll fix the conflicts which I just created now. :-)  Cheers!

jzohrab:
Thanks very much @parradam, super!

jzohrab:
Launched in 3.9.5.

PR #589, opened 2025-03-07
Add AAC, FLAC and WebM support

imamcr:
Allowed popular audio files.


jzohrab:
Thank you very much @imamcr .  Probably the text hint/comment for the form should be in some kind of icon as the list is getting long, but that's for a later PR.  Cheers!

jzohrab:
Launched in 3.10.1.

PR #590, opened 2025-03-08
Add PWA support

imamcr:
Added PWA support.

image
image


jzohrab:
Hi @imamcr - I don't feel that this is sufficient, but I honestly don't know. :-)  To me, "install app" implies that it can be installed with the full db, back end running, etc, and this will be misleading to users.  Thoughts?

imamcr:
> Hi @imamcr - I don't feel that this is sufficient, but I honestly don't know. :-) To me, "install app" implies that it can be installed with the full db, back end running, etc, and this will be misleading to users. Thoughts?

I see your point! However, "install app" in the PWA context simply means adding it to the home screen and running it in its own window, making it feel more like a native app. It doesn't imply installing the backend or database. This is a common way PWAs work, and many users expect this behavior.

Many modern web apps allow users to "install" them as PWAs, giving them a more native-like experience.

image

Lute is in my app launcher Rofi:

image

Users can launch the site from their app launcher, and have an app-like experience without needing a browser tab.

image

jzohrab:
Nice.  @imamcr , how does the PWA work with things like tabs and popups?  (Sometimes Lute opens things up in separate tabs, and some dictionaries have to be opened as a popup.)  -- sorry am pressed for time hence the basic questions :-P

Also, how do I get at the "install app" link, if I install Lute using pip?

imamcr:
1. If you are using an external dictionary, a new window will open. Like a popup. If the "Open popup in new tab" behavior is selected, the dictionary will open as a new tab in the browser.
2. After navigating to localhost:5001, the "Install Lute" icon will appear in the address bar. Click on it, install the app, close the page and launch Lute from an app launcher.

imamcr:
@jzohrab forgot to tag you.

jzohrab:
Thanks @imamcr for the ping.  My only remaining concern for this is that it may cause extra questions for users who see the "install app" thing and think that it runs self-contained, and doesn't need the back end.  I wonder if we need some docs for the manual.

schlick7:
This is pretty old, but many times it is "Add to Home Screen" or "Add to Desktop"

jzohrab:

I think, before this gets merged, that I"d need some docs for the user manual, something including the following:
* launching the PWA doesn't actually launch the back end as well, user still needs to start the python or docker process and leave it running for the PWA thing to actually work
* hmm ... anything else?  Any upgre concerns, weird javascript hiccups?  Don't know!

jzohrab:
Hi again @imamcr - can you make any suggestions for the manual?  e.g., maybe a new page for the PWA support, and then what is needed for it if anything, as I gave in the prior comments.

PR #603, opened 2025-04-06
Add usage heatmap calendar with UTC-based layout and fix stats page

rdearman:
Added a heatmap to the stats page so you can see how consistent you've been about studying. This is like the anki statistic. 

This PR fixes broken stats and heatmap rendering:

    Updates /stats/data to work without wordsread table

    Adds a yearly UTC-based activity heatmap styled like AnkiDroid

    Uses only current year data, auto-adjusting per year

Fully tested on my own fork with real usage data.

Let me know if any changes are needed!


rdearman:
I am going to remove this because I misunderstood the word read part of the dB. I will fix that and redo the request

Issue #178, opened 2024-02-04
Add word stats breakdown


jzohrab:
Add a table of the word counts/term counts by status.

shetozz:
Instead of a table, can we add a pie chart like Anki's

![image](https://github.com/jzohrab/lute-v3/assets/93838524/9de9d892-4728-472d-8a75-412429479041)
**It should show the different stats of terms, and total number of terms, both in numbers and percentages**
*Not sure if you also can track this on a timeline so you can add lines for different term states like the current one for the number of read words.

jzohrab:
Lute currently doesn't track the status hist over time, it just has the current status -- so you won't see this kind of history.  It could be added, that'd be straightforward, and maybe some people would find it motivating.

PR #604, opened 2025-04-07
Add Yearly Activity Heatmap to Stats Page -- INCOMPLETE

rdearman:
This PR introduces a yearly activity heatmap to the /stats page. It visualises user activity over the current year (e.g., 2025), based on the number of WoStatusChanged updates per day.

### Features:
Visual heatmap using a calendar-style layout (similar to AnkiDroid)

Data is pulled from the existing words table, using WoStatusChanged dates

Only shows activity for the current calendar year, dynamically updated

Displays in UTC to ensure consistent date mapping across timezones

Includes a legend showing activity intensity levels

### Notes:
Existing stats functionality is untouched (original /stats/data logic remains)

Heatmap rendering and endpoint are fully self-contained

Let me know if you'd like any changes or enhancements!


jzohrab:
Hi @rdearman - sorry for the delay in responding.  Thanks for the PR!

A heatmap based on word status dates will be misleading, so I think it should be based on words read instead (which is the `wordsread` table, populated when the user clicks the icons at the bottom of each page) -- but that's already kind of implicit in the "words read" chart.

e.g., if I created a bunch of status 1 words yesterday while reading, but today I read more and don't create or update any new words, a heatmap based on word status dates would only show the items from the prior day, even though I read a bunch today.  Another example: if I read 10 new words last week, and then updated their statuses again this week, the heatmap would only show the entries for this week, and nothing for last week.

I think that to do a word-status-based heatmap correctly, Lute would need to track the status history of words, and not just track the current status.  The idea of tracking status history was mentioned in [issue "Add word stats breakdown"](https://github.com/LuteOrg/lute-v3/issues/178), and then maybe this work could be combined with that.

Thoughts?

Cheers and thanks again for taking the time with this submission!

rdearman:
I looked at that. But I am looking at it more like an anki heat map which would show me that I have consistently been using it every day. The problem with words read is that it is only updated when you move to the next page. For me learning Korean I am only reading about 3 sentences a week and it is going to take ages before wordsread gets updated. So while perhaps more accurate it is less motivating. I did consider putting in an activity table and creating a trigger to show when you were active on that day. Might need to rethink it.

jzohrab:

Hi again @rdearman , hope all is well.  There's been a recent "reading streak" PR which _might_ address some of what you want, _maybe_.  Check out #668 and #669 .

jzohrab:
Following up:  I do like the heatmap idea, but to base it on word status alone won't be sufficiently accurate, and since the "page read" doesn't work for you, we're kind of at a loss on how to proceed.  For this to have the complete picture that you need, we need to track the "word status history" as well (ref https://github.com/LuteOrg/lute-v3/issues/178), and then combine this with the "page read" somehow, as that really is what the activity is.

I'm marking the PR as "incomplete" so that someone can come and finish it if they want, including the status history mentioned above.  I won't push for this PR to be finished and merged.  This PR is a good base, and hopefully someone will be motivated.  Cheers and thanks!

jzohrab:
Again, thanks for the PR.  I'm closing it as it is either stale, or not ready to merge in its current form. I've added the "future-work" label because there are ideas and implementation details here that could be useful if someone revisits this feature in the future.

Issue #608, opened 2025-04-11
Encoding issue importing EPUB


tbm:
**Description**

I'm importing an EPUB.  There's an encoding issue with characters.

If I unpack the EPUB manually, `file` says it's a standard UTF-8 file:
```
file Trompie_02.xhtml
Trompie_02.xhtml: HTML document, Unicode text, UTF-8 text, with CRLF line terminators
```
ebook-viewer from Calibre shows it correctly.

**To Reproduce**

Steps to reproduce the behavior, e.g.:

1. Go to 'Books'
2. Click on 'Create new Book'
3. Choose: Language: Afrikaans; Text file: click on epub. Click on 'Save'
4. Open book - see wrong characters

**Screenshots**

Lute:

![Image](https://github.com/user-attachments/assets/118a5103-cb99-45b5-955f-d2ecc979d317)

Calibre:

![Image](https://github.com/user-attachments/assets/d5688e4c-cfd9-41ac-bc7a-44f57755abb9)

**Extra software info, if not already included in the Description:**

- Debian GNU/Linux
- Firefox, but same in Chrome
- Python, pip3 install with venv, according to the instructions
- Version: 3.10.1

tbm:
I noticed the default characters for Afrikaans are `a-zA-ZÀ-ÖØ-öø-ȳáéíóúÁÉÍÓÚñÑ`.  I removed that because it says the default is all UTF-8 characters. After changing this, I deleted the book and imported again but that didn't make a difference.

tbm:
What's strange is that the same character shows up just fine a little bit later:

![Image](https://github.com/user-attachments/assets/a2d156ac-b7a5-4626-9dcf-a56c61656b0f)

Looking at the EPUB, the heading comes from Trompie_02.xhtml while the text is in Trompie_02-4.xhtml. I'm not sure I see any difference. Both are UTF-8 encoded. Both have the same `<body... xml:lang="en-US"` info and there's no meta info about encoding.

jzohrab:
Hmm ... Lute uses [openepub](https://github.com/sakolkar/openepub) for parsing, which was written by @sakolkar -- I wonder if he has any ideas about what's up with this.

sakolkar:
This seems to me as a bug in the openepub library. I'll see if I can figure out a fix. What I suspect is happening is that the UTF-8 file with the headings is being interpreted as ISO-8859-1.


[...]

PR #617, opened 2025-05-15
GH-608: openepub decoding fix

sakolkar:
Bumping openepub library version.

New in 0.0.9 is a bugfix when text match multiple encodings. Openepub will try to use Unicode first and then fallback to other encodings.


jzohrab:
Thank you @sakolkar - I've been busy!  Appreciate the PR, cheers, I'll merge it soon.  :wave:

jzohrab:
Oof, I have totally fallen out of the dev mindset with Lute, due to various life events.  Apologies again @sakolkar :-) I really should get to this soon.

axel584 (approved):

jzohrab:
Launched in 3.10.2.

PR #628, opened 2025-08-05
Remove test dealing with removed demo data

MookLags:
I noticed this TODO under a test that loaded demo data which was I think was removed in Issue 534. It looks like it's unused. I figured that it was small enough that I didn't need to do an issue. Is that alright?

### DIFF
--- a/tests/features/test_rendering.py
+++ b/tests/features/test_rendering.py
@@ -24,13 +24,6 @@ text = None
 
 scenarios("rendering.feature")
 
-
-@given("demo data")
-def given_demo_data(app_context):
-    "Calling app_context loads the demo data."
-    # TODO remove this
-
-
 @given(parsers.parse("language {langname}"))
 def given_lang(langname):
     svc = LanguageService(db.session)


jzohrab (changes requested):
I don't believe this can be removed, as the "Given demo data" is referenced in one of the feature files:

```
cat ./features/rendering.feature
Feature: Rendering
    Text is rendered correctly.

    Background:
        Given demo data
```

If you can confirm that this "@given" can be removed without breaking the acceptance tests, then we should be fine, but I'm reluctant to merge it.

(I should remove the TODO, I'm not even sure why it's there!)

jzohrab:
Thanks @MookLags for submitting, this breaks tests :-) e.g.

```
FAILED tests/features/test_rendering.py::test_rendering_smoke_test - pytest_bdd.exceptions.StepDefinitionNotFoundError: Step definition is not found: Given "demo data". Line 5 in scenario "Rendering smoke test" in the feature "/home/runner/work/lute-v3/lute-v3/tests/features/rendering.feature"
```

Note this isn't a test, it's a step definition used by the feature files.

I'm not sure why the TODO is there .... blah, should have left a better comment, because I would have had a reason as to why something needed to be removed or changed.

jzohrab:
Closing PR, breaks tests.  Unfortunately that TODO is still there ... and I can't recall why. :-/

PR #638, opened 2025-09-17
Return dictionary object instead of URL

### LINKED ISSUE #331: Change dictionary javascript structure used by UI, dicttabs
Currently, the dictionaries are given to the UI in a simple array like this (more or less)

```
[ "http://blah/###", "*http://blarg/###" ]
```

In issue #330, we'd also need to pass along the nickname, presumably something like this:

```
[ { url: "http://blah/###", nickname: "d1" }, { url: "*http://blarg/###", nickname: "d2" } ]
```

So as a prep. move for that, we should change the current dict structure to this:

```
[ { url: "http://blah/###" }, { url: "*http://blarg/###" } ]
```

and while doing that, we might as well get rid of that "*" character:

```
[ { url: "http://blah/###", dicttype: "embedded" }, { url: "http://blarg/###", dicttype: "popup" } ]
```

Sizing notes:

- change existing dictionary structure returned by route
- check everywhere that the dict struct is used
- change dicttabs.js and stuff to work with existing
- ensure sentence translation works (embedded and popup)
- ensure page translation works (embedded? and popup)
- ensure term dict tabs work

goodolcuppa:
Closes #331 

Returns a dictionary object in place of the URL, including both URL and dicttype, removing logic around the prepended '*', and altering the corresponding test.

### DIFF
--- a/lute/models/language.py
+++ b/lute/models/language.py
@@ -25,14 +25,6 @@ class LanguageDictionary(db.Model):
     is_active = db.Column("LdIsActive", db.Boolean, default=True)
     sort_order = db.Column("LdSortOrder", db.SmallInteger, nullable=False)
 
-    # HACK: pre-pend '*' to URLs that need to open a new window.
-    # This is a relic of the original code, and should be changed.
-    # TODO remove-asterisk-hack: remove * from URL start.
-    def make_uri(self):
-        "Hack add asterisk."
-        prepend = "*" if self.dicttype == "popuphtml" else ""
-        return f"{prepend}{self.dicturi}"
-
 
 class Language(
     db.Model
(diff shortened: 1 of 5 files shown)


jzohrab:
Thank you for the PR and I apologies for the inactivity, life and vision issues have caused some troubles.  The code looks good on first glance, super.  I hope to re-review it and merge it early next month.  Thank you again.

jzohrab:
@goodolcuppa -- thank you again for the great PR.  I've long disliked the dictionary "*" hack I had, and this change might open up other possibilities too like dictionary plugins for local offline lookups.  Cheers and regards, much appreciated!

jzohrab:
Launched in 3.10.2.

PR #645, opened 2025-12-05
[Fix] Tibetan renders properly

tomer-gom:
### Summary

Currently Tibetan isn't rendered properly, characters aren't stacking as they are supposed to. This is because the font used doesn't support Tibetan script stacking. This PR addresses that.

- Add a Tibetan font fallback (local-only, unicode-range: U+0F00-0FFF) so Tibetan text stacks correctly without bundling fonts.
- Load that fallback once after /theme/current so it wins over the theme’s body font but still respects user custom styles.
- Preserve each theme’s typography by setting --lute-theme-font-stack to match its existing font stack; To keep each theme’s typography intact (and avoid falling back to the base stack), each theme needed to set --lute-theme-font-stack to match its existing font choices. That way, the new fallback can prepend Tibetan-capable fonts while preserving the look each theme already defined for non-Tibetan text.
- Include the fallback CSS on all pages (reader, term popup, term sentences, image search) so term-related views get the same Tibetan rendering.

### Testing

- Manual: verified Tibetan text stacks in reader, title, term popup/selector, sentences.
- Manual: confirmed script-fallbacks.css loads after /theme/current on those pages.
- Manual negative checks: temporarily broke the Tibetan stack and the theme fallback to ensure the Playwright assertions fail when expected.
- Automated: created pytest tests/playwright/test_tibetan_font.py (with app running on :5001) passes when fallbacks are correct and fails under the simulated breakages (simulated Tibetan text using not supported font, and simulated non-Tibetan text overriden with Tibetan font instead of theme font)

Before (Characters not stacking properly)
Screenshot 2025-12-06 at 0 13 25
After (Correct character stacking)
Screenshot 2025-12-06 at 0 13 02


tomer-gom:
Opened new pr - https://github.com/LuteOrg/lute-v3/pull/647

Issue #236, opened 2024-02-11
Show more detail in hover popup (children, siblings, ...) with options for level of detail


jzohrab:
With some languages and users, it would be nice to see more detail in the hover.  e.g. Modern Greek has tricky verb forms, and being reminded of these on hover could be helpful.

Requested info: parents, children, sibling terms, perhaps just with some non-cluttered styling and info like translation; or even just a list of children/siblings would be good, b/c their data can be derived from the parent.  If a term has multiple parents, the children could be listed under the parent term.

There's a _lot_ of info possible, and could be overwhelming/crowded.  Alternatives:

* add UserSettings for what levels of detail to include in the popup (this is a lot of little bits of code, low ROI?)
* show extra hover detail on certain key combination?  e.g. hover + shift?
* place hovered information in other frames?

GyckaZ:
Additional idea closely related to this one: to be able to click on any of parents and sibling terms within the frame and being directed to that word with its translation, parent terms and sibling terms within the same frame (if it was possible to navigate back, it would be perfect). If not possible within the same frame, then in additional frame - this assuming that the inactive frame (without mouse pointer on it) would be automatically closed after x ms (configured in settings for example).

jzohrab:
Right, I think I'd need someone to investigate js popup libraries to see what's possible. 👍

jzohrab:
Related: https://github.com/jzohrab/lute-v3/issues/341

jzohrab:
https://tippyjs.bootcss.com/ looks good, I think @webofpies did some hacking with this earlier.

webofpies:
my work is in this branch if anyone wants to continue to work on it: https://github.com/webofpies/lute-v3/tree/tippyjs
tippy also supports popup inside popup if needed for parent terms

jzohrab:
Thanks @webofpies 👍  - it feels like this is the right thing to do, do you agree?  -- created issue https://github.com/jzohrab/lute-v3/issues/360 to investigate.

webofpies:
yep. i think it's more future proof with much more cool features

jzohrab:
Marking as blocked by #360 b/c that issue's code is a good starting point.

PR #651, opened 2026-02-11
WIP - Feature/inner terms cycle

tomer-gom:


jzohrab:
Hi @tomer-gom -- sorry for not following up on this earlier, you know the reason :-)

This looks pretty ambitious, but I couldn't quite grok the intent.  I wonder if there is overlap between this and https://github.com/LuteOrg/lute-v3/issues/236 .  In summary, some terms get pretty complicated, and I thought that the popup would let you drill into different components of it if you wanted, parents, children, components, etc.  It looks like this is kind of what you're trying to get to.

Issue #652, opened 2026-02-11
chinese.yabla.com refuses connection when embedded


DorusKeijzer:
i am using the lute3-mandarin plugin, which previously allowed me to look up characters using chinese.yabla.com (the default). Since today this site refuses connection when embedded in lute.

ssoberun:
Can verify that I am experiencing the same issue.

I am on Linux, and this is happening on both Firefox and Chromium.

DorusKeijzer:
oh yes, i'm also on linux on firefox. Perhaps I should have mentioned that in my first commenti

DorusKeijzer:
@ssoberun 
putting https://www.mdbg.net/chinese/dictionary?page=worddict&email=&wdrst=0&wdqb=[LUTE] as dictionary setting works fine as a replacement for me btw

ssoberun:
Thank you!

jzohrab:
(Apologies for the delay in responding, I've had some vision issues, [detail here](https://jzohrab.github.io/lute/2026/04/04/lute-dev-stopped/))

Some sites have been adding security restrictions preventing them from being embedded.  It's quite annoying, and I don't know what the benefit is for them to do so ... but it's out of Lute's control.  Super that you've found a replacement though.  Cheers!

DorusKeijzer:
Sorry to hear about your vision man. I can only imagine what it's like and it sounds horrible to have to go through this. I wish you all the best and hope the situation stabilizes. 

I imagine they think automated/programmatic access would hurt their bottom line. They sell their own subscription service and I reckon they think free language learning tools are competition to them.

ssoberun:
I wish you well jzohrab! I really didn't know, and always put yourself over this wonderful project you've made!
I only wish the best for you. Please take good care of yourself, and thank you so much for this wonderful resource!

jzohrab:
Thank you, https://github.com/LuteOrg/lute-v3/pull/654 has been merged into `develop`, appreciated!

jzohrab:
Launched in 3.10.2.

PR #654, opened 2026-02-22
fix: Changed Mandarin default dictionary from yabla (broken) to MDBG

DorusKeijzer:
In #652, i reported that yabla.chinese.com (the default dictionary in the Mandarin plugin) no longer allows embedding. One other user corroborated that this is true for him as well, so in this pull request i changed the default dictionary to mdbg.net.

### DIFF
--- a/plugins/lute-mandarin/definition.yaml
+++ b/plugins/lute-mandarin/definition.yaml
@@ -2,7 +2,7 @@ name: Mandarin Chinese
 dictionaries:
   - for: terms
     type: embedded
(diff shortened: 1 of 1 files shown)


jzohrab:
Thank you very much @DorusKeijzer , appreciated!

jzohrab:
Marking as in-progress b/c I believe that this will require a plugin release to pypi as well.

jzohrab:
New plug-in released to pypi per instructions at https://github.com/LuteOrg/lute-v3/wiki/Releasing-language-parser-plugins, this is really done now.  Thanks again @DorusKeijzer .

jzohrab:
Launched in 3.10.2.

PR #657, opened 2026-03-07
feat(language): add monolingual mode

shipurjan:
Monolingual mode is for when you're learning a language using that language itself, like using a French dictionary to learn French, rather than translating everything to English.

  When you enable it on a language, the "Translation" field on terms becomes a "Definition" field where you write definitions in the target language itself.

  The main feature is a new "view mode" for Well Known (status 99) words: clicking one in the reading view opens a panel showing the definition you wrote as parsed, interactive reading text, so you can read the definition and click through words in it the same way you would in the main text. From there you can keep drilling into nested definitions, with a back link to retrace your steps. There's also an edit button if you need to modify the card.

image

image

Here's the dockerhub link to the image built from the latest commit if you want to test it:

https://hub.docker.com/r/shipurjan/lute3/tags

### DIFF
--- a/lute/db/schema/migrations/20260307_add_LgIsMonolingual.sql
+++ b/lute/db/schema/migrations/20260307_add_LgIsMonolingual.sql
@@ -1,0 +1,1 @@
+ALTER TABLE languages ADD COLUMN "LgIsMonolingual" TINYINT NOT NULL DEFAULT 0;
--- a/lute/language/forms.py
+++ b/lute/language/forms.py
@@ -48,6 +48,7 @@ class LanguageForm(FlaskForm):
         FormField(LanguageDictionaryForm, default=LanguageDictionary)
     )
     show_romanization = BooleanField("Show Pronunciation field")
+    is_monolingual = BooleanField("Monolingual mode")
     right_to_left = BooleanField("Right-to-left")
 
     # Note!  The choices have to be set in the routes!
--- a/lute/models/language.py
+++ b/lute/models/language.py
@@ -60,6 +60,7 @@ class Language(
     _word_characters = db.Column("LgRegexpWordCharacters", db.String(500))
     right_to_left = db.Column("LgRightToLeft", db.Boolean)
     show_romanization = db.Column("LgShowRomanization", db.Boolean)
+    is_monolingual = db.Column("LgIsMonolingual", db.Boolean)
     parser_type = db.Column("LgParserType", db.String(20))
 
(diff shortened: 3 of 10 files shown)


jzohrab:

Hi @shipurjan -- thank you for the PR, and wow this is an interesting idea -- and I don't know how to get my head around it :-) or how to have it fit within the extremely basic idea of Lute.  What you've done looks pretty slick, from the image, though I haven't tried it out.  In the past I've done hacky things like create an ongoing book (in Lute) where I sometimes copied interesting sentences, and the Lute would parse it for my processing.

I can see how this would be satisfying to use.  On the other hand, it gets into a rabbit hole for learners, where they start to read dictionary entries instead of just continuing reading their books/articles.  From that perspective, I don't feel it fits with Lute's very simple core idea of a fun easy-ish-to-use reading tool.

This feels like a niche use case that would only be used by a small handful of users.  I could be wrong, of course, but given my feeling, I don't want to pull this as it stands into the main Lute development/release line.  I'll assign this back to you so we can continue to chat about it.

shipurjan:
Hej @jzohrab, thanks for replying

Yeah, it does create a possibility for users to get into definition "rabbit holes" but in my opinion it's up to the user how they want to use the software.

It would be quite similar to Wikipedia, where some people might read entire articles, and other people might get into hyperlink rabbit holes.

But I understand that Lute wasn't designed to be used like this from the beginning. I mostly added it for myself to have this "Wikipedia mode", because at a certain level of learning a foreign language you can start reading the definitions in that language as well. But maybe you might not know parts of explanation in that definition. But having a shared pool of words and the same component for both views helps you go into the definition "rabbit holes" until you hit all definitions that you already know.

I agree that it's a niche use case, and don't expect it to be merged. My implementation probably also doesn't cover all edge cases. I just wanted to share it to share an idea for using Lute also for monolingual reading and just use the dictionary for unknown words. Not necessarily for dual-language translation

jzohrab:
Thanks @shipurjan - I'm going to mark this as "wont-merge", but it's such an interesting idea that I'll leave it open, at least for now.  Thank you for posting it though, maybe at some point it will morph into something that I can grok for the necessary docs and discussion, etc!  Cheers.

jzohrab:
@shipurjan -- Thanks again for the PR!  I'm closing it per prior discussion. I've added the "future-work" label because there are ideas and implementation details here that could be useful if someone revisits this feature in the future.  Cheers!

Issue #168, opened 2024-02-01
Add basic "rich-text" features to texts


M-Biggles:
It would be very handy to be able to create headings, indent, align, use bold/italic/underline/etc in the texts we create/read.

jzohrab:
This is a nice idea, and it has the same challenges as #169 (add images) - the parsers would need to distinguish between words and presentation tokens.  Thanks for the issue, cheers.

jzohrab:
I was looking into adding images to texts, which relates to this issue.  The easiest way I can think of to do this would be to use Markdown as the text specification, e.g. `**bold**` is rendered as "**bold**" here is GitHub, etc, but that creates some problems when applied to existing non-markdown texts.

With e.g. MIT-licensed library mistletoe, it's possible to parse markdown.  e.g this document

      # TESTING
      
      **Here** is some stuff.
      
      ```
      python3.8 -m venv .venv
      ```
      
      trying out mistletoe -- [link](https://github.com/miyuchina/mistletoe?tab=readme-ov-file)
      
      Here is some text that is broken up into shorter lines
      but I'm pretty sure that markdown will combine them into one.
      So for book imports, this might not work well.
      Because.
      These might all be paragraphs.
      
      This would def be a separate paragraph.
      
      ---
      
      page 2?
      
      ```
      MacBook-Pro:zz_mistletoe_test jeff$ source .venv/bin/activate
      (.venv) MacBook-Pro:zz_mistletoe_test jeff$ pip3 install mistletoe
      ```

gives this ast (abstract syntax tree)

```
{
  "type": "Document",
  "footnotes": {},
  "line_number": 1,
  "children": [
    {
      "type": "Heading",
      "line_number": 1,
      "level": 1,
      "children": [
        {
          "type": "RawText",
          "content": "TESTING"
        }
      ]
    },
    {
      "type": "Paragraph",
      "line_number": 3,
      "children": [
        {
          "type": "Strong",
          "children": [
            {
              "type": "RawText",
              "content": "Here"
            }
          ]
        },
        {
          "type": "RawText",
          "content": " is some stuff."
        }
      ]
    },
    {
      "type": "CodeFence",
      "line_number": 5,
      "language": "",
      "children": [
        {
          "type": "RawText",
          "content": "python3.8 -m venv .venv\n"
        }
      ]
    },
    {
      "type": "Paragraph",
      "line_number": 9,
      "children": [
        {
          "type": "RawText",
          "content": "trying out mistletoe -- "
[...]

PR #664, opened 2026-04-05
Chapter progress bar

p-value-sharma:
## What this does

Adds a chapter progress bar to the reading view. It uses Lute's existing
bookmarks feature as chapter markers — bookmark the first page of each
chapter and the bar automatically tracks your progress within that chapter.

The bar is hidden when no bookmarks exist before the current page, so it
doesn't clutter the UI for books without chapter markers.

**Example display:**
> Chapter 3 · Page 4 of 19 · 15 pages left · 21%

## How to use

Bookmark the first page of each chapter via the hamburger menu →
Bookmarks → Add Bookmark, naming it e.g. "Chapter 1", "Part II", etc.
The progress bar appears automatically below the audio player while reading.

## Changes

- `lute/bookmarks/routes.py` — new endpoint `/bookmarks/<bookid>/chapter_progress/<pagenum>` that calculates chapter position from bookmarks
- `lute/templates/read/index.html` — progress bar HTML and JavaScript
- `lute/static/css/styles.css` — progress bar styles using existing CSS variables

## Notes

Developed with assistance from Claude (Anthropic). Tested locally.

### DIFF
--- a/lute/bookmarks/routes.py
+++ b/lute/bookmarks/routes.py
@@ -63,6 +63,72 @@ def add_bookmark():
     return jsonify(success=True, status=200)
 
 
+@bp.route("/<int:bookid>/chapter_progress/<int:pagenum>", methods=["GET"])
+def chapter_progress(bookid, pagenum):
+    """
+    Return chapter progress info for the given page.
+
+    Bookmarks are used as chapter markers — the bookmark whose page is <= the
+    current page is the chapter start; the next bookmark's page - 1 is the
+    chapter end (or the last page of the book if there is no next bookmark).
+
+    Returns JSON:
+      { found: false }                          — no bookmarks before this page
+      { found: true, title, pages_in,
+        chapter_length, pages_left, pct }       — chapter info
+    """
+    br = BookRepository(db.session)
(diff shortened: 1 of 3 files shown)


jzohrab:
Hi @p-value-sharma -- I think I'm going to have to decline this one within the Lute release, as people use the bookmarks in various ways, not necessarily conforming to chapter headings.  e.g. I use it sometimes to mark characters, grammar notes, nice passages, etc.

It's still a great idea though.  Lute does need some way of identifying chapters.  Currently bookmarks are just plain data, it's rather clunky.  Having them be "classified" into certain types (chapter heading, eg) would be very natural, but that's a much bigger change than what you've given here.  Some people have also suggested using something like Markdown (see #168) which might then correspond to chapters nicely.  That's a more interesting, and more painful, change.

Back to you so that we can chat about this a bit more if you'd like.  Thanks again, appreciate the effort!

jzohrab:
Thank you again for the PR.  I'm closing it per prior discussion as it's not ready to merge in its current form. I've added the "future-work" label because there are ideas and implementation details here that could be useful if someone revisits this feature in the future.

PR #665, opened 2026-04-05
Fix audio player visibility in Night theme

p-value-sharma:
The audio player uses hardcoded light colors that make the time text invisible in Night theme. Override the player's CSS variables and background colors to match the dark theme.

### DIFF
--- a/lute/themes/css/Night.css
+++ b/lute/themes/css/Night.css
@@ -117,3 +117,44 @@ span.hamburger {
 #reading_menu .reading-menu-item:hover {
   color: var(--background-color);
 }
+
+/* Fix: audio player night theme */
+.audio-player-container {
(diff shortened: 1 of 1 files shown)


axel584:
without the fix : 
image

with the fix : 
image

The fix seems good to me, thank you ! :+1:

axel584 (approved):
The fix seems good to me, thank you !

jzohrab:
Yes, that's really sharp.  Thanks @p-value-sharma for the fix!  (And thank you @axel584 for the images, very helpful 👍 )  Cheers!

PR changed to the `develop` branch, which is where PRs go.  Thanks again!

jzohrab:
Launched in 3.10.2.

PR #666, opened 2026-04-07
Add Restore for Backups

schlick7:
This was coded entirely with an LLM. I tried to keep it to the point and not touch anything not needed. 

I have tested it locally with various backup files from different systems and all seem to work fine. I would recommend further testing by others though to confirm. 

My reason for wanting this is to smooth the transition from going between Termux when I travel and my Server when I'm home. 

### Summary

- Added backup upload support on the backups page
- Added restore and delete actions for backups
- Expanded the backups list to show all `.db.gz` files so uploaded backups are visible (I can switch this to renaming the uploads instead if preferred)
- Uploaded backups are saved with a `manual_` prefix so they are not auto-pruned
- Restore preserves the current server's `backup_dir` after cross-system restores

### Notes

- The `backup_dir` preservation is intended to avoid breaking backups after restoring a DB created on a different system with different folder paths
- Other machine-specific paths are not touched but we may want to add them as well.

### DIFF
--- a/lute/backup/routes.py
+++ b/lute/backup/routes.py
@@ -5,6 +5,7 @@ Backup settings form management, and running backups.
 """
 
 import os
+import re
 import traceback
 from flask import (
     Blueprint,
@@ -51,6 +52,29 @@ def download_backup(filename):
     settings = _get_settings()
     fullpath = os.path.join(settings.backup_dir, filename)
     return send_file(fullpath, as_attachment=True)
+
+
+@bp.route("/upload", methods=["POST"])
+def upload_backup():
+    """
+    Upload a backup file into the backup directory.
+    """
+    upload = request.files.get("backup_file")
+    if upload is None or upload.filename == "":
+        flash("No backup file selected.", "error")
+        return redirect("/backup/index", 302)
+
+    filename = os.path.basename(upload.filename)
+    if not re.match(r"(manual_)?lute_backup_.*\.db\.gz$", filename):
+        flash("Invalid backup filename.", "error")
+        return redirect("/backup/index", 302)
+
+    settings = _get_settings()
+    os.makedirs(settings.backup_dir, exist_ok=True)
+    dest = os.path.join(settings.backup_dir, filename)
+    upload.save(dest)
+    flash(f"Backup uploaded: {filename}", "notice")
+    return redirect("/backup/index", 302)
 
 
 @bp.route("/backup", methods=["GET"])
--- a/lute/backup/routes.py
+++ b/lute/backup/routes.py
@@ -93,6 +93,26 @@ def do_backup():
         return jsonify({"errmsg": str(e) + " -- " + tb}), 500
 
 
+@bp.route("/do_restore", methods=["POST"])
+def do_restore():
+    """
+    Ajax endpoint to restore a backup file.
+    """
+    prms = request.form.to_dict()
+    filename = prms.get("filename", "")
+
+    c = current_app.env_config
+    settings = _get_settings()
+    service = Service(db.session)
+    try:
+        service.restore_backup(c, settings, filename)
+        flash(f"Backup restored: {filename}", "notice")
+        return jsonify(filename)
+    except Exception as e:  # pylint: disable=broad-exception-caught
+        tb = traceback.format_exc()
+        return jsonify({"errmsg": str(e) + " -- " + tb}), 500
+
+
 @bp.route("/skip_this_backup", methods=["GET"])
 def handle_skip_this_backup():
     "Update last backup date so backup not attempted again."
--- a/lute/backup/routes.py
+++ b/lute/backup/routes.py
@@ -18,7 +18,7 @@ from flask import (
 )
 from lute.db import db
 from lute.models.repositories import UserSettingRepository
-from lute.backup.service import Service
+from lute.backup.service import Service, BackupException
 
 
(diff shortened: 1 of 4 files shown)


jzohrab:
Hi @schlick7 - thanks for the interesting PR!  In the past I'd considered doing something like this but wasn't sure if it was a good idea ... data is such a precious thing, and I thought that making restores _too easy_ would somehow be a bad thing :-)  Anyway, I'll add some comments.  Responding to these comments may be out of your range of expertise, in which case we'd want someone to pick it up and run with it.

lute/backup/routes.py
```diff
@@ -53,6 +53,34 @@ def download_backup(filename):
     return send_file(fullpath, as_attachment=True)
 
 
+@bp.route("/upload", methods=["POST"])
+def upload_backup():
+    """
+    Upload a backup file into the backup directory.
+    """
+    upload = request.files.get("backup_file")
```
jzohrab, line 61: Most of this code belongs in the `lute/backup/service.py`.  The controller should remain as thin as possible and delegate things to the service layer.  Then the code should be tested in `tests/unit/test_backup.py`, just to ensure that the code always works when others change it.  Other services throw service-layer exceptions, you can probably find examples in the code.

lute/backup/service.py
```diff
@@ -112,6 +114,34 @@ def create_backup(self, app_config, settings, is_manual=False, suffix=None):
         self._remove_excess_backups(settings.backup_count, settings.backup_dir)
         return f
 
+    def restore_backup(self, app_config, settings, filename):
```
jzohrab, line 117: I feel that this code needs to handle a few different cases, all of which should be really checked in the test_service.py:

* if I upload a .db.gz that was _not_ a lute db, then everything will break.  The code should first check that the db is in fact a valid db.  It would need to connect to the code and then check for at least a few tables and columns (probably not all, just a few things should suffice)
* if the user restores an old db, then the code has to run db migrations on it.  I don't know if you'd need to do a total restart or not ... I don't think so, would need to check the startup code
* if the db is bad, or the migrations fail (?), then the old db should be kept in place, nad the error reported back to the user

This section needs a fair amount of checking before the backup can be automated like this.  At least with the user handling this manually, we can say that it's out of the code's hands, and so don't have to handle it.  If the code manages it, then it has to handle everything.

lute/backup/service.py
```diff
@@ -189,5 +219,22 @@ def list_backups(self, outdir) -> List[DatabaseBackupFile]:
         return [
             DatabaseBackupFile(os.path.join(outdir, f))
             for f in os.listdir(outdir)
-            if re.match(r"(manual_)?lute_backup_", f)
+            if f.endswith(".db.gz")
```
jzohrab, line 222: This should be fine, but it implies that the user won't try to upload/restore a bad db :-)

lute/backup/service.py
```diff
@@ -189,5 +219,22 @@ def list_backups(self, outdir) -> List[DatabaseBackupFile]:
...
         ]
+
+    def _get_backup_file(self, outdir, filename) -> DatabaseBackupFile:
+        "Return a validated backup file."
+        fullpath = os.path.join(outdir, os.path.basename(filename))
+        return DatabaseBackupFile(fullpath)
+
+    def _set_backup_dir(self, dbfilename, backup_dir):
+        "Preserve the local backup_dir after restore."
```
jzohrab, line 231: Given that this refers to settings, it likely belongs in the `lute/settings` somewhere, and then the backup code would refer to that.  This keeps all of the setting logic/structure in the same place.

jzohrab (changes requested):
Overall this is good work (thank you!), but it could create a real problem if a bad upload is done.  This needs a whack of automated tests to ensure that it handles weird cases well.

Given the complexity of the cases that need to be handled, I wonder if this is necessary ... but that's just me, and having a graceful restore would be nice.

lute/backup/service.py
```diff
@@ -112,6 +114,34 @@ def create_backup(self, app_config, settings, is_manual=False, suffix=None):
...
+            with gzip.open(backupfile.filepath, "rb") as in_file, open(
+                tempname, "wb"
+            ) as out_file:
+                shutil.copyfileobj(in_file, out_file)
+
+            bind = self.session.get_bind()
+            self.session.close()
+            self.session.remove()
+            bind.dispose()
+            os.replace(tempname, app_config.dbfilename)
```
jzohrab, line 139: The current db should likely be backed up before it is overwritten/deleted!

jzohrab:
Another thing that I haven't really thought about is how the restore interacts with user media (images) -- right now, the backup is just the db backup, I think, but maybe it needs to somehow be expanded to include media -- and that could be come a space drain.

lute/backup/service.py
schlick7: I just had this thought after your initial comment. This should definitely be done.

schlick7:
I knocked out the couple easy ones. 

Confirming the database seems like it would require a lot of code to be added. Easily done with an LLM, but it will probably be a little messy.

schlick7:
I just uploaded a first pass at a db validation before restoring. Not sure if it goes far enough for you or not.

jzohrab:
I’m away for three weeks, will check on return. This is a very tough feature!  Cheers @schlick7 thanks.

PR #669, opened 2026-04-19
Reading streak functionality

### LINKED ISSUE #668: Reading streak
to boost motivation to read every day

ideally, a heatmap would be nice

could look like the anki review heatmap plugin

#669

VeryUsual:
Fixes #668

### DIFF
--- a/lute/templates/index.html
+++ b/lute/templates/index.html
@@ -41,4 +41,10 @@
   {% include "book/tablelisting.html" %}
 {% endif %}
 
+{% if reading_streak > 0 %}
+<div style="text-align: center; margin-top: 20px; font-size: 1.2em;">
+  <strong>Reading Streak: {{ reading_streak }} day{{ reading_streak != 1 and 's' or '' }}</strong>
+</div>
+{% endif %}
+
 {% endblock %}
--- a/lute/templates/index.html
+++ b/lute/templates/index.html
@@ -41,7 +41,7 @@
   {% include "book/tablelisting.html" %}
 {% endif %}
 
-{% if reading_streak > 0 %}
+{% if reading_streak > 0 and show_streak_on_home %}
 <div style="text-align: center; margin-top: 20px; font-size: 1.2em;">
   <strong>Reading Streak: {{ reading_streak }} day{{ reading_streak != 1 and 's' or '' }}</strong>
 </div>
--- a/tests/unit/stats/test_service.py
+++ b/tests/unit/stats/test_service.py
@@ -104,4 +104,4 @@ def test_get_reading_streak(spanish, app_context):
     assert get_reading_streak(db.session) == 2
 
     make_read_text(spanish, "Otro text.", day_before_yesterday)
-    assert get_reading_streak(db.session) == 2
+    assert get_reading_streak(db.session) == 3
--- a/tests/unit/stats/test_service.py
+++ b/tests/unit/stats/test_service.py
@@ -94,18 +94,14 @@ def test_get_reading_streak(spanish, app_context):
     today = datetime.now().date()
     yesterday = today - timedelta(days=1)
     day_before_yesterday = today - timedelta(days=2)
-    three_days_ago = today - timedelta(days=3)
 
-    assert get_reading_streak(db.session) == 0
-
-    make_read_text(spanish, "Yo tengo un gato.", yesterday)
     assert get_reading_streak(db.session) == 0
 
     make_read_text(spanish, "Ella esta aqui.", today)
     assert get_reading_streak(db.session) == 1
 
-    make_read_text(spanish, "Nuevo text.", day_before_yesterday)
+    make_read_text(spanish, "Nuevo text.", yesterday)
     assert get_reading_streak(db.session) == 2
 
-    make_read_text(spanish, "Otro text.", three_days_ago)
+    make_read_text(spanish, "Otro text.", day_before_yesterday)
     assert get_reading_streak(db.session) == 2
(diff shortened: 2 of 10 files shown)


tests/unit/stats/test_service.py
```diff
@@ -87,3 +87,27 @@ def test_get_data_works_when_nothing_read(app_context):
...
+
+    assert get_reading_streak(db.session) == 0
+
+    make_read_text(spanish, "Yo tengo un gato.", yesterday)
+    assert get_reading_streak(db.session) == 0
+
+    make_read_text(spanish, "Ella esta aqui.", today)
+    assert get_reading_streak(db.session) == 1
+
+    assert get_reading_streak(db.session) == 2
```
jzohrab, line 104: Why would the reading streak increase here, if nothing has happened?

lute/templates/index.html
```diff
@@ -41,4 +41,10 @@
   {% include "book/tablelisting.html" %}
 {% endif %}
 
+{% if reading_streak > 0 %}
```
jzohrab, line 44: I think I'd prefer this to be on the stats page, to keep the home screen as uncluttered as possible (it's already pretty busy!).  I recognize that this is a matter of taste; I personally wouldn't want it :-) I try to model the style on Anki, sort of, and they keep their stats on their stats summary.

Maybe this could be on the stats page always, and then there could be a user setting, "Show streak on home page" ?  Perhaps that's overdesign, but it would let you have what you want (streak on the home screen) and let me hide it for myself.  (Another possibility would be for "home screen widgets", but that is _much more involved!_ and overkill for a free app.)

jzohrab (changes requested):
Looks like lint has failed (`inv lint` will show the problem), and maybe black will fail too.  Also I have the question about the extra assert in the test, and the tests are failing.  But this is all a draft PR anyway!

I think some people like stats like these.  I personally find it hard to say when streaks and stats become a form of punishment :-) , which is why I haven't really added any of them!

Thank you for the clean code and the tests!

jzohrab:
Hi @VeryUsual -- thanks for the PR.  You opened it as draft, so maybe this is just checking if such a feature would be a good add or not.  You can find me on Discord if you wanted to chat about things beforehand.

Side note: I don't like turning contributors away, and on the other hand, we need to try to keep the contributions moving the code forwards ... hard to say what that would be, sometimes, and discussing things would probably help me articulate where I think things need to go.  On the other hand, even though I'm the main writer of Lute, I don't want it to be _my_ project, you know?  It's _ours_.

Some notes in the review above, happy to chat.  Cheers and thanks again!

VeryUsual:
The linting job may have an error, it says 10.00/10 but fails, which is odd.
image

jzohrab:
Hi @VeryUsual, re "The linting job may have an error, it says 10.00/10 but fails ...": it actually shows the errors just above that.  I'm not sure why it shows 10.00/10:

```
************* Module lute.stats.service
lute/stats/service.py (115): Line too long (103/100) (C0301 line-too-long)
************* Module lute.stats.routes
lute/stats/routes.py (017): Line too long (110/100) (C0301 line-too-long)
```

lute/templates/index.html
jzohrab: Hi @VeryUsual , you marked this as resolved but I think it still needs a check of the setting to see if it should be shown on this page, and it should be stats/index.

jzohrab (changes requested):
Thank you @VeryUsual for adding the setting (that setting code is not that great, on review I realize it is pretty confusing at first glance!).

With the lint fix, streak added to stats output, and the setting check on the main page, it should be good to go, unless I missed something!

jzohrab (changes requested):
Looks great! Just need to fix the lines that lint says are too long :-)

```
lute/stats/service.py (115): Line too long (103/100) (C0301 line-too-long)
lute/stats/routes.py (017): Line too long (110/100) (C0301 line-too-long)
```

jzohrab:
Thanks @VeryUsual for the final checks.  The failed tests aren't due to your work, re-running them.  Cheers!

jzohrab:
Super, thanks very much @VeryUsual .  Did you have any last tweaks you wanted to add?  e.g. I see that you didn't include the reading streak in the stats, even though your commit https://github.com/LuteOrg/lute-v3/pull/669/commits/43925eb70ab421da2371e7c5ad3c790c4d1bec9e said you had.

VeryUsual:
Oh, turns out I didn't actually add those files, so they weren't included in the commit. I'll fix that now.

VeryUsual:
Okay. I confirmed by after running a completely new instance cloned from my fork. Everything is now good.

jzohrab:
Looks good, thank you again @VeryUsual for all of the small changes, appreciated.  I'll check this out locally and rebase before I merge it in, but this looks good to be included.  Thank you!

axel584 (approved):
look great for me

jzohrab:
I'm making one small adjustment: I've moved the "reading streak" message on the stats page to below the graph:

image

Reasons for this:
* It keeps the "words read" info all together, vs injecting additional info into the middle of it
* I have a feeling that someone is going to want to add extra information to the streak functionality, i.e. some kind of "heat map".  This idea has come up before by several people, so putting it here means less thrash later if someone eventually wants to add it.

I'm going to rebase the branch locally and squash to a single commit before merging.  Normally I don't do that, but it keeps the commit history cleaner, and this is a well-managed change.

Thank you again @VeryUsual , appreciated!

jzohrab:
Replaced by #676, thank you again!

PR #671, opened 2026-04-19
Add explicit "mimetype" to ensure CSS is handled correctly

### LINKED ISSUE #630: Add explicit "mimetype" to ensure CSS is handled correctly
In issue #623 , @newexploder found and fixed a CSS issue where the files weren't getting loaded correctly.  Weird that it happened on his/her machine only (so far), but the fix should probably be applied to the app code as it's very fast.

Borrowing from the notes in that issue:

---

### Root Cause
The root cause was an incorrect **MIME type** being served for the CSS files.

-   **Expected `Content-Type`:** `text/css`
-   **Actual `Content-Type` served:** `application/x-css`

### The Solution
The issue was resolved by adding two lines of code to the application factory file (`lute/app_factory.py`) to explicitly register the correct MIME type for CSS files within the Python environment.

1.  Add the `mimetypes` import to the top of the file:
    ```python
    import mimetypes
    ```

2.  Add the following line after all the imports, before the `create_app()` function:
    ```python
    mimetypes.add_type('text/css', '.css')
    ```

After adding this code and restarting the server, the CSS is now served with the correct `text/css` MIME type, and the application's layout renders perfectly.

---

This should be a _very_ fast fix.

VeryUsual:
Fixes #630

### DIFF
--- a/lute/app_factory.py
+++ b/lute/app_factory.py
@@ -248,9 +248,9 @@ def _add_base_routes(app, app_config):
         Some files should never be cached.
         """
         response = make_response(send_from_directory("static/js", filename))
-        response.headers[
-            "Cache-Control"
-        ] = "no-store, no-cache, must-revalidate, max-age=0"
+        response.headers["Cache-Control"] = (
+            "no-store, no-cache, must-revalidate, max-age=0"
+        )
         return response
 
     @app.errorhandler(500)
@@ -378,7 +378,9 @@ def _init_parser_plugins(plugin_data_path, outfunc):
     for _, v in supported_parsers():
         outfunc(f"  * {v.name()}")
 
-mimetypes.add_type('text/css', '.css')
+
+mimetypes.add_type("text/css", ".css")
+
 
 def create_app(
     app_config_path=None,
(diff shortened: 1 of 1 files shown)


jzohrab (approved):
Looks good to me, thank you @VeryUsual !

jzohrab:
fyi PRs should go to `develop`, changed it.  Thanks 👍

jzohrab (changes requested):
PR code fails the `black` test (run `inv black`), please push a fix and then things should be good.

jzohrab:
hmm ... again failed ... when I run black locally, I get this change:

```
         response = make_response(send_from_directory("static/js", filename))
-        response.headers["Cache-Control"] = (
-            "no-store, no-cache, must-revalidate, max-age=0"
-        )
+        response.headers[
+            "Cache-Control"
+        ] = "no-store, no-cache, must-revalidate, max-age=0"
         return response
```

maybe due to differences in python version, etc, not your issue @VeryUsual :-) .  Seems trivial, so I'll fix it locally and then will merge.  Cheers and thank you!

jzohrab:
Yeah very odd black behaviour.  Ah well.  Cheers!

jzohrab:
Launched in 3.10.2.

PR #672, opened 2026-04-19
Implement book finished screen with total words read after each book -- INCOMPLETE

VeryUsual:
image

### DIFF
--- a/lute/static/css/styles.css
+++ b/lute/static/css/styles.css
@@ -998,6 +998,99 @@ div.code pre {
     background-image: url("/static/icn/font-decrease.svg");
 }
 
+.book-finished-overlay {
+    position: fixed;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+    background-color: rgba(0, 0, 0, 0.7);
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    z-index: 10000;
+}
+
+.book-finished-content {
+    background-color: white;
+    padding: 40px;
+    border-radius: 10px;
+    text-align: center;
+    box-shadow: 0 4px 6px rgba(0, 0, 0, 0.1);
+    max-width: 500px;
+    animation: slideUp 0.3s ease-in-out;
+}
+
+@keyframes slideUp {
+    from {
+        opacity: 0;
+        transform: translateY(20px);
+    }
+    to {
+        opacity: 1;
+        transform: translateY(0);
+    }
+}
+
+.book-finished-content h1 {
+    font-size: 72px;
+    margin-bottom: 20px;
+}
+
+.book-finished-content h2 {
+    font-size: 36px;
+    margin: 20px 0;
+    color: #333;
+}
+
+.book-finished-content p {
+    font-size: 18px;
+    color: #666;
+    margin: 10px 0;
+}
+
+.book-finished-content .words-read {
+    font-size: 24px;
+    color: #4CBB17;
+    margin: 20px 0;
+    font-weight: bold;
+}
+
+.book-finished-actions {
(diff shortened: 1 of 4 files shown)


axel584 (approved):
It works well on my machine, and I like the visual appearance of the result.

I find the code very clean. Using the service to calculate the word count is very effective.

Great work!

lute/static/css/styles.css
```diff
@@ -998,6 +998,99 @@ div.code pre {
...
+    font-weight: bold;
+}
+
+.book-finished-actions {
+    margin-top: 30px;
+    display: flex;
+    gap: 10px;
+    justify-content: center;
+}
+
```
jzohrab, line 1065: Trying this out locally.

* the .btn and btn-xxxx styles shouldn't be added.  Those classes are added to buttons elsewhere, and these colors don't work in general.  Eg for the language screen, see the screenshot below.

image

lute/templates/read/index.html
```diff
@@ -540,10 +540,35 @@ <h2>&#127881;</h2>
...
 
+  function showFinishedScreen(total_words) {
+    const html = `
+      <div id="book-finished-modal" class="book-finished-overlay">
+        <div class="book-finished-content">
+          <h2>Book Finished!</h2>
+          <p>Great job!</p>
+          <p class="words-read">You've read <strong>${total_words}</strong> words.</p>
+          <div class="book-finished-actions">
+            <a href="/" class="btn btn-primary">Home</a>
```
jzohrab, line 559: Removing the .btn style will make this "home" link to just be a link, not a button :-)  Currently, the last page of the book an "archive" link, which I kind of like having.  I think that just having a "Close" button is sufficient -- or, perhaps best would be to just have the message get put into a div in the page reading page itself, without the popup overlay.  See the notes in the popup overlay css above.

lute/static/css/styles.css
```diff
@@ -998,6 +998,99 @@ div.code pre {
     background-image: url("/static/icn/font-decrease.svg");
 }
```
jzohrab, line 1000: This is good on its own, but these prescribed styles won't work well for some themes.  Some users have specific displays (e.g grayscale) that require certain colors, so setting colors here means that all themes need to be touched as well, which is a drag.

Rather than do a popup, I think that the cleanest and least intrusive method would be to show the "congratulations" message in the simple system font, just as a div within the reading page.  Fewer popups = fewer style issues, and less styling code etc overall.

lute/templates/read/index.html
jzohrab: After thinking through the styling implications, I do feel that a method like `showFinishedMessage` would be best.

jzohrab (changes requested):
@VeryUsual , I'd like the message to be shown in a div within the reading page, for the reasons given above.  If you have time to make those changes, great, and if not I can use the work you've done here to get it in.  Does that make sense?

Cheers and thanks very much!

edit: PS - I've contributed to open source projects before, and sometimes getting pushback like this from the owners feels a bit thankless.  I hope that's not the case here, I appreciate you taking the time to push the project forwards constructively.

jzohrab:
Marking the PR as "incomplete", as it needs changes before it can be merged.  I'll still leave it assigned to @VeryUsual for when you have a moment to make the changes, or perhaps another dev will see this and be motivated to make the changes.  I don't have time or energy at the moment myself.

jzohrab:
Thanks for the PR, I'm closing it due to inactivity ... it's not quite ready to merge in its current form as mentioned above. I've added the "future-work" label because there are ideas and implementation details here that could be useful if someone revisits this feature in the future.

PR #673, opened 2026-04-19
Change submodule URL to use HTTPS

VeryUsual:
Change submodule URL to use HTTPS to ensure easier access for all contributors in areas where SSH keys may not be available

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "lute/db/language_defs"]
 	path = lute/db/language_defs
-	url = git@github.com:LuteOrg/lute-language-defs.git
+	url = https://github.com/LuteOrg/lute-language-defs.git


jzohrab:
Thank you @VeryUsual .  I've never bothered to look into the difference between these, but it does look like https is better for a casual, open team.  Anki uses https as well in their repo (ref https://github.com/ankitects/anki/blob/main/.gitmodules). Devs can override that to ssh if they want -- `git config submodule.<name>.url git@github.com:org/repo.git`

I don't see why the tests failed, I'll rerun the job.

I'll also try out another local clone to see what cred prompts or anything I get, b/c that would need to be documented for new devs.

jzohrab:
Hm repeated failures, though the submodule checkout succeeded in CI.  Not sure why, will try running the dev line in CI again.

jzohrab:
Continues to die after repeat reruns.  I'll have to try this locally to see if I can sort out what's up, as this doesn't make sense that it would fail after a successful checkout.

PR #681, opened 2026-05-10
fix: import web page

EternalDreamer01:
Add User Agent header to avoid forbidden HTTP issue 403.

__Example:__
```
https://es.wikipedia.org/wiki/Felis_catus
```

### DIFF
--- a/lute/book/service.py
+++ b/lute/book/service.py
@@ -176,6 +176,10 @@ class FileTextExtraction:
 
 class Service:
     "Service."
+    
+    HEADERS = {
+        "User-Agent": "Lute/3.0 (X11; Ubuntu; Linux x86_64)"
+    }
 
     def _unique_fname(self, filename):
         """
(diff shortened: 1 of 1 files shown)


jzohrab:
Thank you!  I believe this user agent should work for other OSs like Mac and windows. Will try on my return. Cheers!

jzohrab:
@EternalDreamer01 there is likely a lint failure. See the failed job.

EternalDreamer01:
It seemed the issue was caused by the dict passed inline, I moved it to the `Service` class as the attribute `HEADERS`

lute/book/service.py
```diff
@@ -176,6 +176,10 @@ def _get_vtt_content(self, filename, filestream):
 
 class Service:
     "Service."
+
```
jzohrab, line 179: This shouldn’t be a class attribute as it is only used by the single method.

EternalDreamer01:
I moved it to the function so

jzohrab:
> It seemed the issue was caused by the dict passed inline, I moved it to the `Service` class as the attribute `HEADERS`

I’ll fix it when I get back. Cheers and thank you again!

jzohrab:
Thank you @EternalDreamer01, I merged this locally, so your work is now in the `develop` branch.  Sorry for the delay in getting this in.  I appreciate the fix!  Cheers.

PR #683, opened 2026-05-11
feat!: autodetect language

EternalDreamer01:
Autodetect language.

__Needs these following pip packages:__
```
langdetect
pycountry
```

### DIFF
--- a/lute/book/routes.py
+++ b/lute/book/routes.py
@@ -129,7 +129,7 @@ def new():
     if isinstance(form.text.data, str):
         language_name = pycountry.languages.get(alpha_2=detect(form.text.data)).name
         language_id = LanguageRepository(db.session).find_by_name(language_name)
-        form.language_id.data = language_id.id if form.language_id.data is not None else 0
+        form.language_id.data = language_id.id if language_id is not None else 0
     
     if form.language_id.data == 0:
         # Don't set the current language before submit.
--- a/lute/book/routes.py
+++ b/lute/book/routes.py
@@ -123,12 +123,19 @@ def new():
     form = NewBookForm(obj=b)
     form.language_id.choices = lute.utils.formutils.language_choices(db.session)
     repo = Repository(db.session)
+    form.language_id.data = 0
+    
+    # Attempt autodetect
     if isinstance(form.text.data, str):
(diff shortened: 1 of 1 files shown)


jzohrab:
Hi @EternalDreamer01 — thanks for the PR, but I don’t feel this change is necessary.  Users specify the language when they create a new book, but it is practically zero effort :-) so there’s no pressing need to optimize this path with code changes and new dependencies.  Back to you for discussion. Cheers!

EternalDreamer01:
Thank you @jzohrab for your reply,
When I was trying to debug the code, it appeared to be more convenient as I was often repeating this step.

I understand that adding dependencies can be bothering, and instead of using these dependencies, it could be possible to use the metadata `lang` inside the HTML body.

If I could implement this instead, would you agree with this ?
Cheers !

axel584:
I agree with jz. I'm not comfortable with the idea of ​​adding additional dependencies.

EternalDreamer01:
Hello @axel584 thank you for your reply,
My current commit is using new dependencies, but the solution I was proposing wouldn't rely on new dependencies.

Using the metadata `lang` inside the HTML body.

If I could implement this instead, would you agree with ?
Cheers,

axel584:
Indeed, that would seem to me to be a better solution, but I'll let jz, the project manager, give his opinion.

jzohrab:
Hi @EternalDreamer01 — this has not been an issue for users I’d be inclined to not implement it :-). It’s extra code and edge cases for to only save users a single click, so I think it’s best to not implement this idea.  I hope this doesn’t dampen your enthusiasm for contributing, it’s definitely appreciated.

I’m happy to chat about ideas on discord or on GH issues too.  Having your work get ignored is demoralizing; on the other hand, at this stage of Lute’s lifetime, we need to ensure that contributions are addressing real issues or are substantially moving the project forwards.  I hope you understand. (I think I need to make this statement clearer somewhere!)

EternalDreamer01:
Thank you for your reply @jzohrab ,
It's okay I understand it can seem less important.
It was mainly when I implemented PR #682  that I had to repeat this step many times.
I'm closing this issue then

PR #684, opened 2026-05-11
Update `lute/book/forms.py` to whitelist `m4b` audio files

MilanTodorovic:
Currently the `FileAllowed` method in `lute/book/forms.py` allows only `m4a` audio formats.
Sources online [0] claim that there is no difference between `m4a` and `m4b` , besides the latter having some more metadata for chapters and stuff.
Renaming an `m4b` file to `m4a` works without any issues and can be played back inside Lute.
Now, the only potential issue would be DRM-locked files, but that should be probably left to the user.

This pull requests whitelists `m4b` and adds `m4a` and `m4b` to allowed formats in the `EditBookForm` class.

[0] https://en.wikipedia.org/wiki/MP4_file_format#Filename_extensions: "Audiobook and podcast files, which also contain metadata including chapter markers, images, and hyperlinks, can use the extension .m4a, but more commonly use the .m4b (MPEG-4 Audiobook) extension."

### DIFF
--- a/lute/book/forms.py
+++ b/lute/book/forms.py
@@ -55,8 +55,8 @@ class NewBookForm(FlaskForm):
         "Audio file",
         validators=[
             FileAllowed(
-                ["mp3", "m4a", "wav", "ogg", "opus", "aac", "flac", "webm"],
-                "Please upload a valid audio file (mp3, m4a, wav, ogg, opus, aac, flac, webm)",
+                ["mp3", "m4a", "m4b", "wav", "ogg", "opus", "aac", "flac", "webm"],
+                "Please upload a valid audio file (mp3, m4a/m4b, wav, ogg, opus, aac, flac, webm)",
             )
         ],
     )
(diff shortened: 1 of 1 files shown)


axel584 (approved):

jzohrab:
LGTM as well. I haven’t tried it out yet with an m4b file yet.

marcomachado:
It is also necessary to adjust the `create_new.html` and `edit.html` files in `templates/book`.

I created a PR with this adjust and creating a single list of allowed audio formats
https://github.com/MilanTodorovic/lute-v3/pull/2

jzohrab:
Rebased and merged manually into the `develop` branch, thank you @MilanTodorovic!  This PR is effectively closed.

@marcomachado - your change is great, I'll pull that in directly.  Cheers!

jzohrab:
@marcomachado:  I created https://github.com/LuteOrg/lute-v3/pull/690 to pull in your code, and fixed the code formatting that `inv black` would have thrown.  Thanks, it's a nice little refactor. 👍

jzohrab:
Launched in 3.10.2.

PR #688, opened 2026-06-24
Fix: Hotkey Conflict Clearing Fix Issue #686

### LINKED ISSUE #686: Hotkey conflict state is not cleared after removing conflicting shortcut
**Description**

When a hotkey conflict is resolved by disabling one of the conflicting shortcuts, the conflict validation is not updated correctly. The hotkey field becomes empty, but it remains marked as conflicting (red border), and the Save button stays disabled.

**To Reproduce**

Steps to reproduce the behavior, e.g.:

1. Go to **Settings**.
2. Open **Keyboard Shortcuts**.
3. Assign a hotkey that is already used by another shortcut.
4. Verify that the conflict warning is displayed.
5. Disable one of the conflicting shortcuts by unchecking its checkbox.
6. Observe that the hotkey field is cleared, but it remains highlighted as a conflict and the Save button stays disabled.

**Expected Behavior:**
The conflict state should be cleared automatically when the conflicting shortcut is disabled, allowing the configuration to be saved.

**Actual Behavior:**
The field remains in a conflicted state even though no hotkey is assigned, and saving remains disabled.

**Workaround**:
Assign a different non-conflicting hotkey to the affected shortcut and then disable it again.

**Screenshots**

Conflicted shortcuts (OK)

Image

After deselect checkbox, still showing text field as red
Image

- OS (e.g., iOS, windows): Linux Manjaro (kernel 6.6.141-1-MANJARO)
- Browser (e.g., chrome, safari): Chrome Version 149.0.7827.102 (Official Build) (64-bit)
- How you've installed Lute (Docker, python, source): Python
- Version: 3.10.1

marcomachado:
closes #686

**Description**
bug when disabling a shortcut did not clear the conflict warning from the remaining shortcut inputs or re-enable the Save button.

**Changes**
**shortcuts.html**

1. Updated the toggle_shortcut function (which runs when a keyboard shortcut's "Enabled" checkbox is toggled) to call _check_dups(). This triggers the duplicate-detection logic and updates validation style and button state dynamically.

### DIFF
--- a/lute/templates/settings/shortcuts.html
+++ b/lute/templates/settings/shortcuts.html
@@ -92,6 +92,7 @@
   function toggle_shortcut(el, text_input_id) {
     let newval = el.checked ? "TBD" : "";
     $(`#${text_input_id}`).val(newval);
+    _check_dups();
   }
 
 </script>
--- a/tests/playwright/playwright.py
+++ b/tests/playwright/playwright.py
@@ -269,3 +269,52 @@ def test_playwright():
     "Run playwright with tests."
     with sync_playwright() as sp:
         run(sp)
+
+
+def test_hotkey_conflict():
+    "Test that disabling a conflicting hotkey clears conflict warning."
+    showbrowser = os.environ.get("SHOW", "") == "true"
+    with sync_playwright() as sp:
+        browser = sp.chromium.launch(headless=not showbrowser)
+        context = browser.new_context()
+        page = context.new_page()
+
+        # Load the shortcuts settings page
+        page.goto("http://localhost:5001/settings/shortcuts")
+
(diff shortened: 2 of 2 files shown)


jzohrab:
Thanks @marcomachado !

your playwright test looks good. There were existing hotkey tests in another spot (https://github.com/LuteOrg/lute-v3/blob/develop/tests/acceptance/reading.feature) and it might be better for these tests to move there, but don’t make any changes. I like the .feature tests, but kind of feel that the feature tests should be using playwright behind the scenes, and so maybe some things should change.  Not sure!

jzohrab:
Confirmed locally that the test fails without the fix, and passes with it.  Thank you @marcomachado for your patience.

PR #689, opened 2026-06-25
fix: clear selected word on page change and fix first word navigation…

### LINKED ISSUE #677: When changing pages, selected word isn't the first one
**Description**

When going to the next page (both through the buttons and through keyboard shortcuts), the selected word isn't the first one of the (new) page, but it selects automatically a word in the middle of the page.

I noticed it has some relation to the position of the selected word in the previous page, but I actually couldn't find the exact pattern.

(When I say "selected", I don't mean that it opens the dictionary, but rather that it is the underlined word, from where it will continue if you click the right/left arrow)

**To Reproduce**

1. Go to 'Settings > Languages'
2. Click on 'Load predefined language and sample stories'
3. Click on 'English'
4. Click on 'Tutorial'
5. Right arrow
6. Next page
7. See unexpected behavior: Underlined word is the second

**Extra software info, if not already included in the Description:**- Windows
- Microsoft Edge
- Python Installation
- Version: 3.10.1

marcomachado:
closes #677

**Description:**
What's happening (Problem)
When navigating pages in Lute (e.g., clicking the next/previous page buttons or using hotkeys), the highlighted/underlined cursor word position from the previous page carried over to the new page instead of being reset.

Additionally, starting navigation on a clean page by pressing the right arrow key skipped the very first word (order 0) and selected the second word instead.

**Cause**

1. Timing race condition on page change: The new page content loaded via AJAX contains an inline script that immediately calls parent.reset_cursor_marker(). However, reset_cursor() (which resets the cursor position LUTE_CURR_TERM_DATA_ORDER to -1) was being executed after the new HTML was inserted. This caused the new page's cursor restoration to read the old page's cursor index.

2. Off-by-one check in _move_cursor(): When no word is highlighted, fe_order defaulted to 0. Moving forward checks for words with order > fe_order (order > 0), which evaluated to false for the first word on the page (whose data-order is 0), skipping it.

**Solution**

1. Moved reset_cursor() to execute before injecting the new HTML in goto_relative_page inside lute/templates/read/index.html
2. Updated the default value of fe_order in _move_cursor inside lute/static/js/lute.js to default to -1 when moving forward (so 0 > -1 matches the first word) and Infinity when moving backward.

**Testing / Verification**

Added an automated Playwright test test_page_change_first_word in tests/playwright/playwright.py verifying that:

1. Moving cursor to a word in the middle of page 1, then changing pages, properly resets the active highlight.
2. Pressing the right arrow key on the new page correctly highlights the first word (order 0 / class kwordmarked) rather than starting on the second word.

### DIFF
--- a/tests/playwright/playwright.py
+++ b/tests/playwright/playwright.py
@@ -29,6 +29,7 @@ Menu sub-items are only visible after hovering over the menu, e.g.:
 
 import os
 import time
+import re
 from playwright.sync_api import Playwright, sync_playwright, expect
 
 
@@ -273,7 +274,7 @@ def test_playwright():
 
 def test_page_change_first_word():
     "Test that going to the next page resets the cursor, so the first word is selected next."
-    import re
+
     showbrowser = os.environ.get("SHOW", "") == "true"
     with sync_playwright() as sp:
         browser = sp.chromium.launch(headless=not showbrowser)
@@ -296,16 +297,20 @@ def test_page_change_first_word():
         page.keyboard.press("ArrowRight")
 
         # Verify a word is highlighted/active
-        expect(page.locator("span.word.wordhover, span.word.kwordmarked").first).to_be_visible()
+        expect(
+            page.locator("span.word.wordhover, span.word.kwordmarked").first
+        ).to_be_visible()
 
         # Go to the next page using the right arrow navigation button
         page.get_by_text("▶").click()
-        
+
         # Wait for the new page content to load
         page.locator("span.word").first.wait_for()
 
         # Verify that NO word on the new page is highlighted automatically
-        expect(page.locator("span.word.wordhover, span.word.kwordmarked")).to_have_count(0)
+        expect(
+            page.locator("span.word.wordhover, span.word.kwordmarked")
+        ).to_have_count(0)
 
         # Press the right arrow key
         page.keyboard.press("ArrowRight")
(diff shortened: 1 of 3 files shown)


jzohrab:
Thank you @marcomachado for the fix!  It's a small fix but still nice to have.  Running it through CI now ... this just needs some lint fixes (and maybe "black" formatting fixes?) and then it should be good to go.

tests/playwright/playwright.py
```diff
@@ -269,3 +269,50 @@ def test_playwright():
     "Run playwright with tests."
     with sync_playwright() as sp:
         run(sp)
+
+
```
jzohrab, line 274: Lint job https://github.com/LuteOrg/lute-v3/actions/runs/28170245507/job/83458566328?pr=689 is giving some things to change:

```
************* Module tests.playwright.playwright
tests/playwright/playwright.py (303): Trailing whitespace (C0303 trailing-whitespace)
tests/playwright/playwright.py (276): Import outside toplevel (re) (C0415 import-outside-toplevel)
```

jzohrab (changes requested):
Small changes, also don't forget to check `inv black` for code formatting.

jzohrab:
Thanks @marcomachado for your patience.  I'm fixing the merge conflicts.

jzohrab:
Replaced by https://github.com/LuteOrg/lute-v3/pull/709, which has the merge conflicts fixes.  Cheers and thanks again!

PR #690, opened 2026-06-25
Unified audio extensions

jzohrab:
Stealing the code from https://github.com/MilanTodorovic/lute-v3/pull/2

Putting it in a PR for visibility and to run through CI.

Thanks to @marcomachado!


jzohrab:
Launched in 3.10.2.

PR #687, opened 2026-06-24
Fix: Resolve broken Bing Image Search Issue #685

marcomachado:
Credits to ignacio (discord ignacio400). He proposed this fix on Discord. I just submitted here

**Description**
Fixes #685

Bing recently updated their HTML layout and bot detection, causing the image search panel in Lute to return few or no images. Three separate issues contributed to this breakage:

1. Wrong CSS classes: Lute filtered images by classes (rms_img and vimgld) which Bing no longer uses.
2. Missing User-Agent: Without a UA header, Bing flags the request as a bot and serves a stripped-down HTML page.
3. Suboptimal Endpoint: Lute was scraping /images/search (a full HTML page that loads images via JS). Since Lute doesn't execute JavaScript, it only saw the sparse initial DOM.

**Changes Proposed**

1. Update Endpoint: Switched the request url to /images/async, which serves image data directly.
2. Add Headers: Added a standard browser User-Agent to bypass the bot-blocker.
3. Parse JSON Data: Changed the parsing logic to extract image URLs from Bing's internal JSON data format (m="{...}" attribute) instead of relying on fragile CSS class names.
4. UI Improvement (Thumbnail Grid): Updated index.html to display the returned images as a compact thumbnail grid (display: flex, 120x90px, object-fit: cover). Previously, images loaded at their full original size, which required excessive scrolling.

PR #691, opened 2026-06-26
Fix bing search

jzohrab:
Replacement PR for #687.

The code here used the code in that PR (in its initial state), with some lint fixes, and then an additional fix as some images weren't saveable (and were throwing errors in the js console when clicked), e.g.:

image

### DIFF
--- a/lute/bing/routes.py
+++ b/lute/bing/routes.py
@@ -49,6 +49,7 @@ def bing_search_page(langid, text, searchstring):
 def bing_search(langid, text, searchstring):
     "Do an image search."
 
+    # pylint: disable=too-many-locals
     # Searching for images slows acceptance tests.  If NO_BING_IMAGES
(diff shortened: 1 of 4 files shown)


jzohrab:
Hi @marcomachado -- I'm pretty sure everything is here.  I took the code prior to your last forced push, so maybe you'd advise making further adjustments.  The biggest add was the last commit to handle some blocked images, as about half of the images were causing me problems.  Thanks!

marcomachado:
> Hi @marcomachado -- I'm pretty sure everything is here. I took the code prior to your last forced push, so maybe you'd advise making further adjustments. The biggest add was the last commit to handle some blocked images, as about half of the images were causing me problems. Thanks!

I see! Seems ok to me!
I didn't see this forbidden error. Thanks for solve it!

jzohrab:
Launched in 3.10.2.

PR #697, opened 2026-07-01
Add hotkey to copy selected words.

### LINKED ISSUE #566: Add hotkey: copy hovered word
Like the title says.  Currently can shift-drag to copy a range, but not a single word.

kblissett:
closes #566

This PR adds a configurable hotkey for copying the selected word(s). When non-contiguous words are selected they will be copied to the clipboard with a single space separating them. This may not be ideal for languages that do not separate words with spaces, but I'm not sure what else to do for the general case of copying non-contiguous spans.

### DIFF
--- a/lute/settings/hotkey_data.py
+++ b/lute/settings/hotkey_data.py
@@ -9,6 +9,8 @@ import yaml
 # Hotkeys and descriptions.
 _ALL_HOTKEY_DATA = """
 Copy:
+- hotkey: hotkey_CopySelected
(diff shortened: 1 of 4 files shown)


jzohrab:
Thanks very much @kblissett for adding the acceptance test, it’s so important!

ci will check lint and code formatting as well so if something comes up, please adjust. Ping me on discord if needed. 👋

kblissett:
Looks like this was clean on CI as far as I can tell. No changes made.

PR #699, opened 2026-07-05
Fix: reuse external dictionary popup window

### LINKED ISSUE #622: how to prevent pop-up windows spamming?
hey man, thank you for this piece of software, already have 13k records. extremely helpful. 

i just wanna ask how to do what the title is talking about? most of online dicts are non-embeddable so i have to select pop-up type for them. but on every new access there come up new window and so i end up with tens of browser windows and laggign system as a consequence. is it possible to make lute re-using existing subwindows instead?

### LINKED ISSUE #698: Feat: Reuse a single external window for popup dictionaries
When using external dictionaries configured as Pop-up window (such as Google Translate, DeepL, or SpanishDict), Lute currently opens a new browser window or tab every time a lookup is performed or a new word is clicked. This quickly results in a large number of open windows that the user must manually close at the end of a reading session.

**Describe the solution you'd like**

This feature unifies and reuses a single external window for all external dictionary lookups (both sentence and term translations) during a reading session; so when an user clicks on a word, Lute should create a new window to lookup meaning, only if the translation window is not already open; otherwise, should reuse existing window and redirect to do a new search

When user closes book (eg.: go to homepage), all translation windows should be closed.

marcomachado:
closes #698
closes #622

This feature unifies and reuses a single external window for all external dictionary lookups (both sentence and term translations) during a reading session; so when an user clicks on a word, Lute should create a new window to lookup meaning, only if the translation window is not already open; otherwise, should reuse existing window and redirect to do a new search

When user closes book (eg.: go to homepage), all translation windows should be closed.

### DIFF
--- a/lute/static/js/dict-tabs.js
+++ b/lute/static/js/dict-tabs.js
@@ -238,7 +238,19 @@ class DictButton extends LookupButton {
     let settings = 'width=800, height=600, scrollbars=yes, menubar=no, resizable=yes, status=no'
     if (LUTE_USER_SETTINGS.open_popup_in_new_tab)
       settings = null;
-    window.open(lookup_url, 'otherwin', settings);
+    
+    const topWin = window.top;
+    if (topWin.lute_popup_window && !topWin.lute_popup_window.closed) {
+      topWin.lute_popup_window.location = lookup_url;
+      topWin.lute_popup_window.focus();
+    } else {
+      const pop = topWin.open(lookup_url, 'dictwin', settings);
+      topWin.lute_popup_window = pop;
+      if (!topWin.lute_popup_windows) {
+        topWin.lute_popup_windows = [];
+      }
+      topWin.lute_popup_windows.push(pop);
+    }
   }
 
   _load_frame(dicturl, text) {
--- a/lute/static/js/lute.js
+++ b/lute/static/js/lute.js
@@ -836,7 +836,19 @@ let show_translation_for_text = function(text) {
     let settings = 'width=800, height=600, scrollbars=yes, menubar=no, resizable=yes, status=no';
     if (LUTE_USER_SETTINGS.open_popup_in_new_tab)
       settings = null;
-    window.open(url, 'dictwin', settings);
+    
+    const topWin = window.top;
+    if (topWin.lute_popup_window && !topWin.lute_popup_window.closed) {
+      topWin.lute_popup_window.location = url;
+      topWin.lute_popup_window.focus();
+    } else {
+      const pop = topWin.open(url, 'dictwin', settings);
+      topWin.lute_popup_window = pop;
+      if (!topWin.lute_popup_windows) {
+        topWin.lute_popup_windows = [];
+      }
+      topWin.lute_popup_windows.push(pop);
+    }
   }
   else {
     top.frames.wordframe.location.href = url;
--- a/lute/static/js/lute.js
+++ b/lute/static/js/lute.js
@@ -836,7 +836,7 @@ let show_translation_for_text = function(text) {
     let settings = 'width=800, height=600, scrollbars=yes, menubar=no, resizable=yes, status=no';
     if (LUTE_USER_SETTINGS.open_popup_in_new_tab)
       settings = null;
-    window.open(finalurl, 'dictwin', settings);
+    window.open(url, 'dictwin', settings);
   }
   else {
     top.frames.wordframe.location.href = url;
--- a/lute/templates/read/index.html
+++ b/lute/templates/read/index.html
@@ -328,6 +328,19 @@
       $('div.audio-player-container').show();
     }
 
+    window.addEventListener("beforeunload", function() {
+      if (window.top.lute_popup_windows) {
+        window.top.lute_popup_windows.forEach(function(w) {
+          if (w && !w.closed) {
+            w.close();
+          }
+        });
+      }
+      if (window.top.lute_popup_window && !window.top.lute_popup_window.closed) {
+        window.top.lute_popup_window.close();
+      }
+    });
+
     goto_relative_page(0, true);
   });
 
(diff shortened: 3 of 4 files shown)


jzohrab:
Wow @marcomachado, this is super, thanks!  I haven't tried it out yet but the code looks good.  Cheers!

tests/acceptance/test_dict_popup.py
```diff
@@ -0,0 +1,83 @@
...
+def test_dictionary_popup_reuses_window(luteclient):
+    """
+    Verify that subsequent lookups reuse the same popup dictionary window.
+    """
+    # 1. Update Spanish dictionary to be popuphtml
+    luteclient.visit(
+        "dev_api/execsql/UPDATE%20languagedicts%20SET%20LdType%20%3D%20%27popuphtml%27"
+    )
+
+    # 2. Create a book in Spanish
```
jzohrab, line 15: Hi @marcomachado -- these comments look like they were AI-generated, right?  Not a problem, but AI can often over-generate comments.  It's not necessary to add a comment when the code is so clear, so you can remove those in future.  Don't bother to change this PR :-)

I'll add a few more comments here, just for your reference.

tests/acceptance/test_dict_popup.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Wait for the dictionary button in the main page and click it
+    dict_btn = luteclient.page.locator(".dict-btn-external").first
+
+    # 5. Click the dict button and expect a popup window to open
+    with luteclient.page.expect_popup():
+        dict_btn.click()
+
+    # Check that a new page was opened
+    assert len(context.pages) == initial_page_count + 1
```
jzohrab, line 33: @marcomachado  - asserts can actually take messages that provide useful error info, so rather than write a comment and then an assert with no info (which would fail with "0 != 1" or similar), you can combine them like this:

```
assert len(context.pages) == initial_page_count + 1, "new page was opened"
```

If this failed, it would say something like "fail: New page was opened (0 <> 1)" or similar, so it provides more information at test failure time.

tests/acceptance/test_dict_popup.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # 6. Now click the word "Adios" in the reading pane to switch to editing "Adios"
+    luteclient.click_word("Adios")
```
jzohrab, line 36: @marcomachado - The code already says what it does, the comment is just repeating that :-)  Redundant = bad, so just the line of code is good.

ok, I think that's enough of my feedback.  Happy to chat on discord if it's helpful.

lute/static/js/dict-tabs.js
```diff
@@ -238,7 +238,19 @@ class DictButton extends LookupButton {
...
+      topWin.lute_popup_window.location = lookup_url;
+      topWin.lute_popup_window.focus();
+    } else {
+      const pop = topWin.open(lookup_url, 'dictwin', settings);
+      topWin.lute_popup_window = pop;
+      if (!topWin.lute_popup_windows) {
+        topWin.lute_popup_windows = [];
+      }
+      topWin.lute_popup_windows.push(pop);
+    }
```
jzohrab, line 253: @marcomachado - This code and the other sections are good, but my concern is that it introduces a concept (the `lute_popup_window` etc) that is spread across a few files, and so it's hard to see how things are connected, or easy to miss.  It would be nice if these three tightly-related things I don't think there's a good way to manage this, though.

lute/static/js/lute.js
```diff
@@ -836,7 +836,19 @@ let show_translation_for_text = function(text) {
...
+      topWin.lute_popup_window.focus();
+    } else {
+      const pop = topWin.open(url, 'dictwin', settings);
+      topWin.lute_popup_window = pop;
+      if (!topWin.lute_popup_windows) {
+        topWin.lute_popup_windows = [];
+      }
+      topWin.lute_popup_windows.push(pop);
+    }
   }
```
jzohrab, line 852: The above block of code is duplicated from dict-tabs.js ... duplication is bad, it would be nice if there was a single source.

dict-tabs.js is included in a few places:
* lute/templates/term/_form.html
* lute/templates/read/index.html

... On the other hand, the js overall is kind of messy, and it's hard to sort out the inclusions neatly, so maybe we just live with this, but add a comment to the code to say that this is duplicate code from dict-tabs.js, and add a TODO comment to say remove the duplication.

lute/templates/read/index.html
```diff
@@ -328,6 +328,19 @@ <h2>&#127881;</h2>
       $('div.audio-player-container').show();
     }
 
+    window.addEventListener("beforeunload", function() {
```
jzohrab, line 331: If we can't somehow move all of the js functions introduced in this PR into a single file source to clarify the connection between the different parts, then this code should have a comment saying something like `top.lute_popup_windows and top.lute_popup_window are set in dict-tabs.js and lute.js`

jzohrab (changes requested):
Hi @marcomachado - a few ponts for discussion here.  I don't know if there's an easy way to centralize the dictionary management stuff to a single location -- the functions could go into a new `lute-popup.js` file which is included in the `base.html` and also in the `templates/term/_form.html`  That way, the management of the popups would be fairly clear.

In addition, we should clean up the excess comments in the acceptance test :-)

Let me know what you think, we can also chat on Discord.  Cheers!

jzohrab (approved):
@marcomachado - I think I'll implement the js idea as documented in recently opened issue #708 -- I'll take care of that, it's quick, and with your tests here it gets exercised.  So I'll merge this and take it back from you.  Thank you very much!

lute/static/js/lute.js
jzohrab: I'll try a quick fix for this, per my comment.  Thanks!

PR #707, opened 2026-07-19
fix: changing image filename exported to anki to use only lowercase

### LINKED ISSUE #701: [Bug] Image filename case mismatch in Anki export (AnkiConnect)
Bug report by @eastwood on [Discord](https://discord.com/channels/1074759089051160647/1525020308627525723)

When Lute exports a term to Anki via AnkiConnect, the  tag written into the note's Image field uses an uppercase filename (e.g. LUTE_TERM_88.jpg), but the actual image file saved to collection.media (via storeMediaFile) is lowercase (e.g. lute_term_88.jpg).
On case-sensitive filesystems like those comonly found in *nix-like systems, this mismatch means Anki can't find the file, so every exported image shows as broken/missing in the Anki card — even though the file exists correctly in collection.media.

Steps to reproduce:
Export a term with an image from Lute to Anki via AnkiConnect
In Anki, open the note in Browse, toggle HTML (shift+ctrl+x) view on the Image field
Note the `` filename is uppercase
Check collection.media the actual file is lowercase. Probably found in ~.local/share/Anki2/User\ 1/collection.media
Card shows broken image icon

Expected: The filename casing used in the `` tag should exactly match the filename Lute passes to storeMediaFile.
Im using Artix Linux, Anki desktop with AnkiConnect plugin and the lastest Lute version as of july 10 2026

marcomachado:
Closes #701

This PR changes the Anki export image filename prefix from uppercase LUTE_TERM_ to lowercase lute_term_ to ensure compatibility with newer versions of Anki.

- Updated handle_image in lute/ankiexport/field_mapping.py to generate the image prefix in lowercase (lute_term_). Both the HTML reference and the stored filename parameter will now be sent to Anki in lowercase.
- Updated unit and integration tests to verify the lowercase prefix and swapped payload execution order.

### DIFF
--- a/lute/ankiexport/field_mapping.py
+++ b/lute/ankiexport/field_mapping.py
@@ -112,7 +112,7 @@ def get_values_and_media_mapping(term, sentence_lookup, mapping):
             ]
             image_srcs = []
             for t, imgfilename in id_images:
-                new_filename = f"LUTE_TERM_{t.id}.jpg"
+                new_filename = f"lute_term_{t.id}.jpg"
                 image_url = f"/userimages/{t.language.id}/{imgfilename}"
                 media_mappings[new_filename] = image_url
                 image_srcs.append(f'<img src="{new_filename}">')
--- a/lute/static/js/lute-anki.js
+++ b/lute/static/js/lute-anki.js
@@ -175,9 +175,9 @@ const LuteAnki = (function() {
    * e.g.
    * [
    *   // This is the result for "export_1"
-   *   ["LUTE_TERM_1.jpg", 1738896726062],
+   *   ["lute_term_1.jpg", 1738896726062],
    *   // This is the result for "export_2"
-   *   ["LUTE_TERM_2.jpg", { "result": null, "error": "... duplicate" }]
+   *   ["lute_term_2.jpg", { "result": null, "error": "... duplicate" }]
    * ]
    *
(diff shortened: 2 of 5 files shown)


jzohrab:
Does it still work with older versions of Anki?  Ie if I use this new code
but I have an older version of Anki, will Anki still show the image?

marcomachado:
Yes! I think so, but, in fact, I didn't test it 

Anki now defines a default way to saved images filenames instead of rely it to other applications (like Lute which decided to use upper case) 

But old cards are already exported and configured using upper case, both in card (img src) and image file in filesystem

If a user uses an older version of anki and creates a new card: Anki will rely of Lute to define filename for images and a card will be generated using lowercase

marcomachado:
Tests:

For new files:
 - ideal scenario: LUTE (branch version) + Anki (current version) = LUTE will create cards already with lowercase; anki will save with lowercase (default)
 
 Alternatives scenarios:
 - [BUG] LUTE (current version) + Anki (current version) = bug (LUTE will use uppercase; Anki now uses lowercase)
 - [OK] LUTE (current version) + Anki (old version) = Anki will rely on Lute to define filename, so files will be saved using uppercase
 - [OK] LUTE (branch version) + Anki (old version) = Anki will rely on Lute to define filename, so files will be saved using uppercase
 
For old files: Anki don't change already saved cards and files

Issue #710, opened 2026-07-21
Refactor js code introduced in pull request #699


jzohrab:
* Refactor pull request #699 js code per ideas in #708 
* Small cleanups of testing code comments

jzohrab:
Merged into develop.

PR #711, opened 2026-07-23
Issue 710: refactor js

jzohrab:
For issue #710

### DIFF
--- a/lute/static/js/lute-popups.js
+++ b/lute/static/js/lute-popups.js
@@ -1,0 +1,42 @@
+/* Lute popups. */
+
+const LutePopups = (function() {
+
+  /* Open a new popup, or reuse existing. */
+  function open_popup(url, settings) {
+    const topWin = window.top;
+
+    if (topWin.lute_popup_window && !topWin.lute_popup_window.closed) {
+      topWin.lute_popup_window.location = url;
+      topWin.lute_popup_window.focus();
+    }
+    else {
+      const pop = topWin.open(url, 'dictwin', settings);
+      topWin.lute_popup_window = pop;
+      if (!topWin.lute_popup_windows) {
+        topWin.lute_popup_windows = [];
+      }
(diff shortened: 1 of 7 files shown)


lute/static/js/lute-popups.js
```diff
@@ -0,0 +1,42 @@
...
+    const topWin = window.top;
+
+    if (topWin.lute_popup_window && !topWin.lute_popup_window.closed) {
+      topWin.lute_popup_window.location = url;
+      topWin.lute_popup_window.focus();
+    }
+    else {
+      const pop = topWin.open(url, 'dictwin', settings);
+      topWin.lute_popup_window = pop;
+      if (!topWin.lute_popup_windows) {
```
jzohrab: @marcomachado - I'm not clear why we need the `lute_popup_windows` array, since there's only one popup open at a time.  Thoughts?

jzohrab:
Just a question to @marcomachado -- otherwise this PR should be good to go.

lute/static/js/lute-popups.js
marcomachado: > @marcomachado - I'm not clear why we need the `lute_popup_windows` array, since there's only one popup open at a time. Thoughts?

This part it's not necessary, that was my mistake. When I first started on this task, I was trying to reuse one popup per external dictionary (which would require tracking them in an array). Later, I changed it to use a single external popup (`dictwin`), but forgot to clean up the array tracking logic.

I will update the logic to only track the active window (`topWin.lute_popup_window`) and remove the redundant `lute_popup_windows` array.

jzohrab: @marcomachado - no problem, I actually have the branch open already on my machine, it's a quick fix.  I'll take care of it, thanks!

jzohrab: Not sure how I missed it in the initial code review, my mistake on that side anyway.  Cheers!

PR #712, opened 2026-07-24
fix: resolve MeCab parser cache invalidation causing 1.5s slowdown on…

chengxicx:
The JapaneseParser.is_supported() cache was being invalidated on every call because mecab_path from settings was empty but _old_mecab_path stored the auto-detected path, causing path_unchanged to be False.

Each page load accessed language.parser ~85 times (once per token), and each call created a new MeCab() instance (~18ms each), totaling 1.5 seconds of avoidable overhead.

Fix: reuse previously auto-detected path when settings path is empty, ensuring the cache stays valid across repeated calls. Also add auto-detection of common MeCab library paths for macOS/Linux.

Performance: 1.5s -> 60ms per page load (25x improvement).

### DIFF
--- a/lute/parse/mecab_parser.py
+++ b/lute/parse/mecab_parser.py
@@ -45,6 +45,18 @@ class JapaneseParser(AbstractParser):
 
         mecab_path = current_settings.get("mecab_path", "") or ""
         mecab_path = mecab_path.strip()
+
+        # If the saved path doesn't exist, try to find a working one.
+        if mecab_path and not os.path.exists(mecab_path):
+            mecab_path = ""
+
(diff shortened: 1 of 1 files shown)


jzohrab:
Thanks @chengxicx -- this sounds like a good contribution, cheers!  It's odd that I didn't have this issue when I was trying out my Japanese ... but things happen.  I had a question around the auto-detect stuff.

bemitc:
Maybe it would be worth it to split the final validation (via calling MeCab()) into an inner function so you can check the validity in the path fallback list when you walk the list.

That way, you handle the possible (but unlikely) edge case where the file exists but it is corrupt, for a different architecture, or for some other reason doesn't work.

Issue #231, opened 2024-02-11
Use ISO 639-3 and ISO 15924 codes


GrimPixel:
**Is your feature request related to a problem? Please describe.**

It is mentioned at https://github.com/jzohrab/lute-v3/issues/116 that a plug-in system and potentially many more stuff need to be done with volunteer support. Therefore a standarisation on the names is needed.

**Describe the solution you'd like**

Use ISO 639-3 and ISO 15924 to indicate language and writing system.

jzohrab:
Thanks for the note.  Can you provide good links for the above?  I'm also not sure where this info should be used.

GrimPixel:
https://iso639-3.sil.org/code_tables/639/data
https://www.unicode.org/iso15924/iso15924-codes.html

jzohrab:
Hi @GrimPixel -- belated follow-up :-)

I'm considering adding an ISO 639-3 field to the languages table so that would be tracked (no deadline just yet), as this could be useful for some text processing work.  Do you think that this is sufficient for the languages table, as a start?

PR #713, opened 2026-07-24
TTS  auto translate

chengxicx:
TTS Speech Synthesis & Automatic Translation
image

Major expansion of Lute3’s text-to-speech functionality:
Edge-TTS Integration: High-quality voice synthesis powered by Microsoft Edge-TTS, Edge browser recommended. 
Page/sentence reading.

On-the-fly Auto Translation: Automatically translates terms into the target language, with a 3-tier fallback chain: Google Translate → MyMemory → backend translation service
image

Settings : hover pronunciation, click-to-read, player display, speaker button visibility
image

Sidebar Controls: Two independent toggles added to the reader sidebar to separately show/hide the TTS player and sentence speaker buttons
image

Unified Play Control: Consolidates the separate play/pause/stop buttons into a single toggle button for play/stop
Playback State Tracking: Introduces an isPaused state variable for reliable playback status monitoring, addressing inconsistent pause detection in the browser’s native SpeechSynthesis API

### DIFF
--- a/lute/read/routes.py
+++ b/lute/read/routes.py
@@ -10,6 +10,7 @@ from lute.term.routes import handle_term_form
 from lute.settings.current import current_settings
 from lute.models.book import Text
 from lute.models.repositories import BookRepository, LanguageRepository
+from lute.tts.routes import get_lang_code
 from lute.db import db
 
 
@@ -36,6 +37,7 @@ def _render_book_page(book, pagenum, track_page_open=True):
         page_count=book.page_count,
         show_highlights=show_highlights,
         lang_id=lang.id,
+        lang_code=get_lang_code(lang.name),
         track_page_open=track_page_open,
         term_dicts=term_dicts,
     )
--- a/lute/read/service.py
+++ b/lute/read/service.py
@@ -156,6 +156,7 @@ class Service:
         "Get paragraphs, set text.start_date if needed."
         text = dbbook.text_at_page(pagenum)
         text.load_sentences()
+
         svc = StatsService(self.session)
         svc.mark_stale(dbbook)
 
@@ -170,6 +171,7 @@ class Service:
         lang = text.book.language
         rs = RenderService(self.session)
         paragraphs = rs.get_paragraphs(text.text, lang)
+
         self._save_new_status_0_terms(paragraphs)
 
         return paragraphs
--- a/lute/settings/current.py
+++ b/lute/settings/current.py
@@ -42,6 +42,12 @@ def refresh_global_settings(session):
         "term_popup_show_components",
         "use_ankiconnect",
         "show_streak_on_home",
+        "tts_enabled",
+        "tts_hover_pronunciation",
+        "tts_click_pronunciation",
+        "tts_show_control_panel",
+        "tts_show_sentence_buttons",
     ]
     for k in boolkeys:
-        current_settings[k] = current_settings[k] == "1"
+        if k in current_settings:
+            current_settings[k] = current_settings[k] == "1"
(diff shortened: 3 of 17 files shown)


lute/read/routes.py
```diff
@@ -36,6 +37,7 @@ def _render_book_page(book, pagenum, track_page_open=True):
         page_count=book.page_count,
         show_highlights=show_highlights,
         lang_id=lang.id,
+        lang_code=get_lang_code(lang.name),
```
jzohrab, line 40: This won't necessarily work ... we can't guarantee that the language name will actually be a language name that resolves to a code.  We may need to release a separate thing where users fill in the language code manually, and then this feature would depend on if that code is set or not.

lute/read/service.py
```diff
@@ -156,6 +156,7 @@ def _get_reading_data(self, dbbook, pagenum, track_page_open=False):
         "Get paragraphs, set text.start_date if needed."
         text = dbbook.text_at_page(pagenum)
         text.load_sentences()
+
```
jzohrab, line 159: Avoid making changes where not needed :-)  It makes reviews harder.  (yes, this is just a blank line!)

lute/settings/current.py
```diff
@@ -33,7 +33,7 @@ def refresh_global_settings(session):
     for h in hotkeys:
         current_hotkeys[h.value] = h.key
 
-    # Convert some ints into bools.
+    # Convert some string values into bools.
```
jzohrab, line 36: I think these are ints, not string values.  Am I mistaken?

lute/settings/routes.py
```diff
@@ -47,20 +47,31 @@ def edit_settings():
         # Update the settings in the database
         for field in form:
             if field.id not in ("csrf_token", "submit"):
-                repo.set_value(field.id, field.data)
```
jzohrab, line 50: I don't follow why this change is needed, as the code handled bools etc correctly before, or so I thought.  Was I mistaken?

lute/tts/routes.py
```diff
@@ -0,0 +1,232 @@
...
+import os
+import asyncio
+import hashlib
+import urllib.parse
+
+import requests
+from flask import Blueprint, current_app, send_file, jsonify
+
+try:
+    import edge_tts
```
jzohrab, line 19: This will only work for MS edge?  What about me on Chrome or others on Firefox?

lute/tts/routes.py
```diff
@@ -0,0 +1,232 @@
...
+    "ca": "ca-ES-JoanaNeural",
+    "bg": "bg-BG-BorislavNeural",
+    "hr": "hr-HR-GabrijelaNeural",
+}
+DEFAULT_VOICE = "en-US-AvaNeural"
+
+# Mapping from the Lute Language "name" to an ISO 639-1 code used by the
+# TTS and translate routes.  Used by read/routes.py to pass the source
+# language code to the reading template.
+LANG_NAME_TO_CODE = {
```
jzohrab, line 66: This isn't a reliable mapping.

jzohrab (changes requested):
While I like the idea of this PR, there are too many variables and gotchas, I think, in the implementation.  I feel that we need users to indicate the language code for their languages, and not try to take a guess.  To that end, I feel that this work is blocked until something like https://github.com/LuteOrg/lute-v3/issues/231 is implemented.

lute/tts/routes.py
chengxicx: Edge browser supports natural voice tones for selection, while other browsers only provide low-quality robotic mechanical voices.
You may test the actual effect on my server.

PR #714, opened 2026-07-24
Backup restore

chengxicx:
Adds a complete end-to-end backup and restore workflow:
image

One-click backup trigger available on the Settings page; archives include the database, configuration files, and book library data
Supports rolling back the system state from any historical backup snapshot
Preserves all user custom preferences (backup directory path, MeCab binary path, etc.) during restoration, preventing overwrites from outdated configs stored in backups
Implements SQLite connection pool reset logic to maintain valid database connections after swapping database files during restore

### DIFF
--- a/lute/app_factory.py
+++ b/lute/app_factory.py
@@ -333,6 +333,40 @@ def _create_app(app_config, extra_config):
     app.db = db
 
     _add_base_routes(app, app_config)
+
+    # The before_request hook checks a single boolean flag.
+    # The import is at module level so it only happens once.
+    from lute.backup.service import Service as _BackupServiceClass
+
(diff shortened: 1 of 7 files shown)


jzohrab:
Thanks for the PR @chengxicx .  There was already a PR open for this at https://github.com/LuteOrg/lute-v3/pull/666, what is different about this PR?

Restoring backups is such a massive thing that I'm not sure how much I _want_ to automate it :-)  I don't know if there will be primary key issues with any things like audio files, images, etc.  This may require a lot of work.

chengxicx:
Thanks, I didn't notice PR #666. The functionality should be similar, so I'll close this PR.

PR #719, opened 2026-07-24
fix: last read time not updating on home page

chengxicx:
The "Last Read" timestamp on the home page does not update immediately after reading. This PR resolves the issue.

- Add lightweight /read/update_start_date endpoint for beforeunload sync
- Add Service.update_start_date method for fast start_date commit
- Track AJAX completion with luteStartReadingDone flag in read page
- Fire synchronous XHR on beforeunload if start_reading AJAX incomplete
- Add pageshow event to refresh DataTables on bfcache navigation


chengxicx:
Need more debug, close.

PR #720, opened 2026-07-24
Fix/last read time

chengxicx:
---
## Pull Request Title
fix: resolve Last Read timestamp not updating & UTC timezone offset issue

### Problem Overview
Two critical bugs exist for the "Last read" column on the home page:
1. **Timestamp fails to update after reading**
After opening a book’s reader page and returning to the home page, the Last Read time stays unchanged.
- Root cause 1: The original update logic used synchronous XHR inside the `beforeunload` event, which modern browsers block or abort when the page unloads.
- Root cause 2: The `luteStartReadingDone` flag blocked all subsequent update requests once the page finished loading.

2. **8-hour timezone offset for UTC+8 users**
Production servers run under UTC. The backend used `datetime.now()` to store server-local UTC timestamps, while the frontend parsed raw timestamps as the browser’s local time via `dayjs(dt)`. This created an 8-hour delay for users in UTC+8 time zones — for example, a book just finished reading would show "8 hours ago".

#### 1. Replace synchronous XHR with `navigator.sendBeacon`
`lute/templates/read/index.html` (Lines 299–314)
- `sendBeacon` is a built-in browser API built for reliable background submission during page unload, with no navigation blocking.
- Adjust the `update_start_date` route to accept both `GET` and `POST`, since `sendBeacon` always sends POST requests.

#### 2. Remove `luteStartReadingDone` guard, add 5-second request throttle
`lute/templates/read/index.html` (Lines 296–302)
- Old logic: No more timestamp updates after initial page load.
- New logic: Trigger updates on every `visibilitychange` (tab hidden) and `beforeunload`, throttled to 5 seconds to prevent redundant API calls.

#### 3. Add `visibilitychange` event listener
`lute/templates/read/index.html` (Lines 316–318)
Guarantees correct timestamp updates for scenarios like switching browser tabs, minimizing windows, or putting mobile browsers into background.

#### 4. Unified timezone handling: store UTC on backend, convert to local time on frontend
- Backend (`lute/read/service.py`, Line 86): Replace `datetime.now()` with `datetime.utcnow()` to save standardized UTC timestamps to database.
- Frontend (`lute/templates/book/tablelisting.html`, Line 526): Append `Z` to timestamp string: `dayjs(dt + 'Z')`. This tells Day.js to parse the value as UTC and automatically convert it to the visitor’s local browser timezone.

#### 5. Fix incomplete refresh logic on `pageshow`
`lute/templates/book/tablelisting.html` (Lines 167–176)
- Original implementation only refreshed data when `e.persisted` (bfcache navigation) was true; normal page navigation back home skipped table refresh entirely.
- Added a one-time load flag to ensure DataTables refreshes for all return-to-home navigation paths.

### Modified Files
| File | Modification Summary |
|------|----------------------|
| `lute/read/routes.py` | Make `update_start_date` endpoint accept POST requests |
| `lute/read/service.py` | Use `datetime.utcnow()` to generate UTC timestamps |
| `lute/templates/read/index.html` | Integrate sendBeacon, visibility listener and 5s throttle |
| `lute/templates/book/tablelisting.html` | UTC timezone parsing fix + full refresh on pageshow |

 
---

### DIFF
--- a/lute/app_factory.py
+++ b/lute/app_factory.py
@@ -62,6 +62,7 @@ from lute.settings.routes import bp as settings_bp
 from lute.themes.routes import bp as themes_bp
 from lute.stats.routes import bp as stats_bp
 from lute.cli.commands import bp as cli_bp
+from lute.tts.routes import bp as tts_bp
 
 
 def _setup_app_dir(dirname, readme_content):
@@ -333,6 +334,7 @@ def _create_app(app_config, extra_config):
     app.db = db
 
     _add_base_routes(app, app_config)
+
     app.register_blueprint(language_bp)
     app.register_blueprint(anki_bp)
     app.register_blueprint(book_bp)
@@ -349,6 +351,7 @@ def _create_app(app_config, extra_config):
     app.register_blueprint(themes_bp)
     app.register_blueprint(stats_bp)
     app.register_blueprint(cli_bp)
+    app.register_blueprint(tts_bp)
     if app_config.is_test_db:
         app.register_blueprint(dev_api_bp)
 
--- a/lute/db/management.py
+++ b/lute/db/management.py
@@ -108,6 +108,12 @@ def add_default_user_settings(session, default_user_backup_path):
         # Anki:
         "use_ankiconnect": False,
         "ankiconnect_url": "http://127.0.0.1:8765",
+        # TTS:
+        "tts_enabled": True,
+        "tts_hover_pronunciation": True,
+        "tts_click_pronunciation": True,
+        "tts_show_control_panel": True,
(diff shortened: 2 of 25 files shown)


jzohrab:
Hi @chengxicx -- this PR contains code from several different fixes, and so can't be considered for merging.  If I merge this, I'm merging everything done before it, so that is a big risk from the code review standpoint.  See the guidelines at https://github.com/LuteOrg/lute-v3/wiki/Contributing#do-one-thing for notes.

If you can pull out the commits with cherry-pick or similar and put them into a separate PR, that would be possible to look it.  As it is, I'm going to close this one.  Cheers and thanks.

Issue #4, opened 2023-05-14
Increase acceptance test coverage (master list)


jzohrab:
Commit ce50ee5112f27 added a basic acceptance (browser-level) test of Lute using Panther: reading a text, and creating Terms and multi-word Terms.

Per https://github.com/jzohrab/lute/blob/develop/tests/acceptance/README.md#tests-to-write, there are a bunch of tests to write, and if extensive work is done on any section of Lute then some of these acceptance might be useful.

- Languages
  - List languages
  - Create new lang
- Create text
  - ~~from textbox~~
  - from file
- Import web page
- Texts
  - archive text
  - view archive
  - unarchive text
  - delete text
- Terms
  - ~~list terms~~
  - search for terms
  - ~~create new term from main form~~
  - bulk map parents from listing
- Term Tags
  - ~~list all~~
  - ~~create~~
  - ~~delete~~
- Reading
  - Update refreshes multiple terms
  - ~~Update on one page updates other books~~
- Parent term mapping
  - export book file
  - export language file
  - import mapping file
- Create backup
- Version and software info

jzohrab:
v3 has `tests/playwright` which do a smoke test run, that should be sufficient for a lot of this.

jzohrab:
Some items cleared thanks to #723

PR #723, opened 2026-08-02
Added four new acceptance tests RE Issue 4

TomStudies:
Referring to this issue: https://github.com/LuteOrg/lute-v3/issues/4

Added four new acceptance tests aimed at addressing the following:

- Languages: List languages
- Texts: Delete text (book)
- Terms: Search for terms
- Backups: Backup setting defaults

These are done through modifications to the existing book and term feature files as well as a new mgmt feature file (I wasn't sure where else to put the language and backup ones) as well as modifications to conftest.py and lute_test_client.py.

inv all passes.

### DIFF
--- a/tests/acceptance/lute_test_client.py
+++ b/tests/acceptance/lute_test_client.py
@@ -106,6 +106,26 @@ class LuteTestClient:  # pylint: disable=too-many-public-methods
 
         self.page.click("#submit")
 
+    def get_language_table_content(self):
+        "Get language table content."
+        self.visit("/")
+        self.page.hover("#menu_settings")
+        self.page.locator("#lang_index").first.click()
+        rows = self.page.locator("#languagetable tbody tr")
+        rowcount = rows.count()
+        content = []
+
+        for i in range(rowcount):
+            row = rows.nth(i)
+            tds = row.locator("td")
+            tdcount = tds.count()
+            rowtext = [tds.nth(j).inner_text().strip() for j in range(tdcount)]
+            ret = "; ".join(rowtext).strip()
+            ret = ret.replace("\u200B", "").replace("\n", "").replace("\\n", "")
+            content.append(ret)
+
+        return "\n".join([c for c in content if c.strip() != ""]).strip()
+
     ################################3
     # Books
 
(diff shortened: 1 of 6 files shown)


jzohrab:
Thank you @TomStudies !  Running CI on it now, if any issues pop up you should get an email.

tests/acceptance/lute_test_client.py
```diff
@@ -379,6 +412,53 @@ def _delimited_tags(td):
         # print(f"RAW ROWSTRINGs = {rowstring}", flush=True)
         return "\n".join([r for r in rowstring if r.strip() != ""]).strip()
 
+    def get_filtered_term_table_content(self):
```
jzohrab: Hi @TomStudies - most of this code is a copy of the code in `get_term_table_content(self):` -- duplication in code is bad, so is it possible to extract a helper function here to get the content?  e.g. something like `_get_term_table_content(self, rows)` or similar, and then both methods could call that.

jzohrab (changes requested):
A request to remove some duplication, if you have time.  If you don't, let me know and I should be able to do it quickly too.  Cheers and thank you!

tests/acceptance/lute_test_client.py
TomStudies: Hi @jzohrab, sorry for the inconvenience! I totally see your point. I can get this done but not right away (I am on vacation and not sure when I will have time until August 16th or so.) I will tackle it then if there is no rush. :)

TomStudies: So sorry for the delay! I have just made this change. inv full passes again.

jzohrab:
Thanks @TomStudies for the revisions, I've been travelling so I'm just looking now.  I've skimmed it and it looks good so far.

tests/acceptance/term.feature
```diff
@@ -6,6 +6,15 @@ Feature: Creating and managing terms
...
             No data available in table
 
+    Scenario: I can search for a term in the term index
+        Given a new Spanish term:
+            text: gato
+            translation: cat
+        When I set the term table filter to "gato"
+        Then the filtered term table contains:
+            ; gato; ; cat; Spanish; ; New (1)
+        When I clear the term table filter
```
jzohrab, line 16: Any reason for this final `When`?  Usually there's a `Then` after one.

jzohrab (approved):
This all looks good, except for that final `When` as noted ... but that's no reason to block merging this.  Thank you very much @TomStudies !

tests/acceptance/term.feature
jzohrab: I'll remove that final `when`, it's not necessary.  Thank you!

jzohrab:
Slight tweaks to the tests in #745 , fyi @TomStudies only.  Thanks again!

